PR #2, opened 2021-02-22
Regenerate with aws-sdk-go v1.35.5

surajkota:
Commands used to generate code

```
export AWS_SDK_GO_VERSION=v1.35.5
export SERVICE=sagemaker
make build-ack-generate
make build-controller SERVICE=$SERVICE
```

We will ignore the AWS SDK version update in the `go.mod` until it is decoupled. We plan to update it before we cut any release and want to keep the scope of this PR to migration

Updated the generator config to ignore new resources so as to reduce a step when we update the sdk version

tested locally using pytest


jaypipes (approved):
Yep, all good IMHO.

RedbackThomson (approved):

PR #3, opened 2021-02-24
Introduce endpointconfig and endpoint controller

surajkota:
### Description of Changes
- Add controllers for endpoint config and endpoint resources
- Following is the specialized code related to endpoint and a brief explanation for each:
   - `custom_set_output`: modifies the `Status/Spec` of the CR after create/describe/update operation
     - This PR adds functionality to modify the `Status.Conditions` and set `TempOutofSync` condition to true unless endpoint is in `InService` or `Failed` state so the reconcile does a periodic requeue until a stable state is achieved
   - `custom_update_api`: intercept methods that trigger an update on an endpoint
     - This PR adds functionality to do a periodic requeue until the endpoint is not in `InService` or `Failed` state
   - `custom_update_conditions`: update conditions based on the status of resource which is not an AWS API error. e.g. Endpoint being in `Failed` status is a terminal state for the controller.
     - This PR adds a check for the example stated above. In future will add more functionality like when an update fails etc.
- keep the resources in ignore list for which controllers exist as commented out for easy reference in future
  - ignore UpdateTrainingJob API for now since it is a special case, only allows modifying few fields and not compatible with the current code-gen pattern

### Testing
Integ test: [aws-controllers-k8s/community#699](https://github.com/aws-controllers-k8s/community/pull/699)
Unit test for custom code: will do in a separate PR

### Notes
Can we merged only after:
- [aws-controllers-k8s/runtime#6](https://github.com/aws-controllers-k8s/runtime/pull/6)
- [aws-controllers-k8s/code-generator#23](https://github.com/aws-controllers-k8s/code-generator/pull/23)

cc [Sagemaker Developer Preview](https://github.com/aws-controllers-k8s/community/projects/3)


go.mod
```diff
@@ -14,3 +14,5 @@ require (
 	k8s.io/client-go v0.18.2
 	sigs.k8s.io/controller-runtime v0.6.0
 )
+
+replace github.com/aws-controllers-k8s/runtime => /home/ubuntu/go/src/github.com/aws-controllers-k8s/runtime
```
surajkota, line 18: remove this and update after [runtime PR](https://github.com/aws-controllers-k8s/runtime/pull/6)

generator.yaml
```diff
@@ -5,13 +5,68 @@ operations:
...
+        is_printable: true
+        from:
+          operation: DescribeEndpoint
+          path: FailureReason
+      EndpointConfigName:
+        is_read_only: true
+        from:
+          operation: DescribeEndpoint
+          path: EndpointConfigName
+      # CreationTime:
```
surajkota, line 55: TODO: add these after delta branch PR is merged to code generator

go.mod
A-Hilaly: You could also keep it in a `go.local.mod` just like elasticache controller => https://github.com/aws-controllers-k8s/elasticache-controller/blob/main/go.local.mod

pkg/resource/endpoint/custom_update_conditions.go
```diff
@@ -0,0 +1,71 @@
...
+		ko.Status.Conditions = append(ko.Status.Conditions, terminalCondition)
+
+	}
+	return true
+}
+
+//toStringPtr returns pointer to a string
+func toStringPtr(str string) *string {
+	return &str
+}
```
RedbackThomson, line 71: Surprised this isn't in common runtime code already?

jkuruba (approved):
Looks good

pkg/resource/endpoint/custom_update_conditions.go
surajkota: replacing this with `aws.String
`

pkg/resource/endpoint/custom_set_output.go
```diff
@@ -0,0 +1,106 @@
...
+
+// customSetOutput sets ConditionTypeResourceSynced condition to True or False
+// based on the endpoint status on AWS so the reconciler can determine if a
+// requeue is needed
+func (rm *resourceManager) customSetOutput(
+	r *resource,
+	endpointStatus *string,
+	ko *svcapitypes.Endpoint,
+) {
+	if endpointStatus != nil {
```
RedbackThomson, line 73: nit: instead do an early return
```go
if endpointStatus == nil {
  return
}
```

pkg/resource/endpoint/custom_set_output.go
```diff
@@ -0,0 +1,106 @@
...
+		syncConditionStatus := corev1.ConditionUnknown
+		switch *endpointStatus {
+		case svcsdk.EndpointStatusInService, svcsdk.EndpointStatusFailed:
+			syncConditionStatus = corev1.ConditionTrue
+			break
+		default:
+			syncConditionStatus = corev1.ConditionFalse
+		}
```
RedbackThomson, line 81: Guessing you expected there to be more conditions? Otherwise this could just be an `if-else`

RedbackThomson (approved):

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/aws-controllers-k8s/sagemaker-controller
 go 1.14
 
 require (
-	github.com/aws-controllers-k8s/runtime v0.0.3
+	github.com/aws-controllers-k8s/runtime v0.0.4-0.20210224005142-96355b6c73d7
 	github.com/aws/aws-controllers-k8s v0.0.2
```
jaypipes, line 7: This needs to be removed.

pkg/resource/endpoint/custom_update_api.go
```diff
@@ -0,0 +1,65 @@
...
+// the autogenerated code that trigger an update on an endpoint
+
+package endpoint
+
+import (
+	"context"
+	"errors"
+
+	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
+	"github.com/aws/aws-controllers-k8s/pkg/requeue"
```
jaypipes, line 24: This needs to change to `github.com/aws-controllers-k8s/runtime/pkg/requeue`

jaypipes (changes requested):
I identified the issue you were having with RequeueNeededAfter being caused by remaining references to `github.com/aws/aws-controllers-k8s/pkg/requeue`. Please see inline.

Also, in the PR message, it states that this PR depends on:

> aws-controllers-k8s/code-generator#23

which has been merged, but I don't actually see anything in this patch that requires a custom sdkFind or sdkDelete method?

pkg/resource/endpoint/custom_update_api.go
surajkota: good catch!!

surajkota:
@jaypipes 
> I don't actually see anything in this patch that requires a custom sdkFind or sdkDelete method?

here are the related methods which needed that PR.
generator.yaml
```
operations:
  ...
  DescribeEndpoint:
      set_output_custom_method_name: customDescribeEndpointSetOutput // defined in custom_set_output.go
  DeleteEndpoint:
      custom_implementation: customDeleteEndpoint // defined in custom_update_api.go
 
```

surajkota:
Ready to merge @jaypipes @RedbackThomson

RedbackThomson (approved):

PR #7, opened 2021-03-19
Endpoint resource: address failed update scenarios

surajkota:
### Description of Changes
- Handle scenarios related to:
  1. **Failing update**. The update endpoint operation is async and describing endpoint only contains information about the current endpoint config. If an update fails, the endpoint goes back to `InService` state with a `FailureReason` about the error. Currently, the controller will be stuck in a loop to keep applying the desired endpoint config and keep on failing with a little window for the customer to notice that the update failed because the failure reason only stays if the endpoint is in `InService` state and disappears as soon as a new update is applied. 
      - This PR fixes the above problem by recording the endpointconfigName used for update under `Status.LastEndpointConfigNameForUpdate` and only retry an update if the `FailureReason` was related to an internal server error or if the user has submitted a new endpointConfigName. Additionally, it sets a `ACK.Terminal `condition to `True` and stops reconciling if either of these conditions is not met. User is expected to check the `failureReason` and re-submit the resource with a **new** endpointConfigName if they wish to update the endpoint. 
      - This PR also adds `Status.LatestEndpointConfigName` so the user can check the current endpoint config name. This is needed in addition to `Spec.endpointConfigName`(desired) for them to have full visibility over the current state of the endpoint
  1.  **Endpoint goes to `Failed` state**. Once the endpoint is in `Failed` state. The only option for the user is to delete and recreate the endpoint.
      -  This PR checks if the endpoint is in `Failed` state as a pre-check-in sdkUpdate and blocks the update. Additionally, it sets `ACK.Terminal` condition to True so that the controller stops reconciling.
  - see `pkg/resource/endpoint/custom_*.go` for related implementation details of the above two scenarios
- Enable retainAllVariantProperties for updates by default since we expect most customers to use this
- Remove tags field from all the resources because these are only created but we don't have an update path. [Tagging experience across ACK](https://github.com/aws-controllers-k8s/community/issues/63) is being worked on separately
- upgrade to latest code-generator and runtime - v0.0.5 and refactor code to use Delta functionality

Code-generator is now adding description to the fields in CRDs and hence the PR looks long but actual changes which need to be reviewed closely are listed above

### Testing
[aws-controllers-k8s/community#730](https://github.com/aws-controllers-k8s/community/pull/730)

Dependent PR: [aws-controllers-k8s/code-generator#34](https://github.com/aws-controllers-k8s/code-generator/pull/34)


pkg/resource/endpoint/custom_update_api.go
```diff
@@ -19,21 +19,59 @@ package endpoint
...
-// customUpdateEndpoint adds specialized logic to requeueAfter until endpoint is in
-// InService state before an updateEndpoint can be called
+var (
+	FailUpdateError = fmt.Errorf("Unable to update Endpoint. Check FailureReason")
+)
+
+// customUpdateEndpoint adds specialized logic to check if controller should
+// proceeed with updateEndpoint call.
+// Update is blocked in the following cases:
+//  1. until EndpointStatus != InService
```
RedbackThomson, line 37: nit: `s/until/while/`

pkg/resource/endpoint/custom_update_api.go
```diff
@@ -19,21 +19,59 @@ package endpoint
...
+// Method returns nil if endpoint can be updated, otherwise error depending on above cases
 func (rm *resourceManager) customUpdateEndpoint(
 	ctx context.Context,
 	desired *resource,
 	latest *resource,
-	diffReporter *ackcompare.Reporter,
+	delta *ackcompare.Delta,
 ) (*resource, error) {
-	return nil, rm.endpointStatusAllowUpdates(ctx, latest)
+	latestStatus := latest.ko.Status.EndpointStatus
```
RedbackThomson, line 47: could early exit here with `if latestStatus == nil: return nil, nil`

pkg/resource/endpoint/custom_update_api.go
```diff
@@ -19,21 +19,59 @@ package endpoint
...
+	lastEndpointConfigForUpdate := desired.ko.Status.LastEndpointConfigNameForUpdate
+
+	// Case 2 - EndpointStatus == Failed
+	if (latestStatus != nil && *latestStatus == svcsdk.EndpointStatusFailed) ||
+		// Case 3 - A previous update to the Endpoint with same endpointConfigName failed
+		// Following checks indicate FailureReason is related to a failed update
+		// Note: Internal service error is an exception for this case
+		// "Request to service failed" means update failed because of ISE and can be retried
+		(latestStatus != nil && failureReason != nil && lastEndpointConfigForUpdate != nil &&
+			!strings.HasPrefix(*failureReason, "Request to service failed") &&
```
RedbackThomson, line 68: move magic string to const?

pkg/resource/endpoint/custom_update_conditions.go
```diff
@@ -51,20 +53,23 @@ func (rm *resourceManager) customUpdateConditions(
...
-		if failureReason != nil {
-			terminalCondition.Message = failureReason
+	if (err != nil && err == FailUpdateError) || (latestStatus != nil && *latestStatus == svcsdk.EndpointStatusFailed) {
+		// setting terminal condition since controller can no longer recover by retrying
+		if terminalCondition == nil {
+			terminalCondition = &ackv1alpha1.Condition{
+				Type: ackv1alpha1.ConditionTypeTerminal,
+			}
+			ko.Status.Conditions = append(ko.Status.Conditions, terminalCondition)
 		}
```
RedbackThomson, line 64: Need an else statement to find the existing terminal condition (if it exists but is set to ConditionFalse)?

pkg/resource/endpoint/custom_update_conditions.go
```diff
@@ -51,20 +53,23 @@ func (rm *resourceManager) customUpdateConditions(
...
-
-		ko.Status.Conditions = append(ko.Status.Conditions, terminalCondition)
-
+		terminalCondition.Status = corev1.ConditionTrue
+		if *latestStatus == svcsdk.EndpointStatusFailed {
+			terminalCondition.Message = aws.String("Endpoint status: Failed. Cannot be updated")
```
RedbackThomson, line 67: nit: This message feels a little clunky? Maybe `Cannot update endpoint with Failed status`?

RedbackThomson (changes requested):

pkg/resource/endpoint/custom_update_conditions.go
surajkota: Good question!
 1. If a terminal condition is already set we do not want to override that, see #51-54
 1. If an error did not occur, this part of code will never execute and the condition will be cleared by [this part of code](https://github.com/aws-controllers-k8s/sagemaker-controller/blob/main/pkg/resource/endpoint/sdk.go#L343-L346). Hence the else does not exist

pkg/resource/endpoint/custom_update_api.go
```diff
@@ -19,21 +19,65 @@ package endpoint
...
+
+	// Case 2 - EndpointStatus == Failed
+	if (latestStatus != nil && *latestStatus == svcsdk.EndpointStatusFailed) ||
+		// Case 3 - A previous update to the Endpoint with same endpointConfigName failed
+		// Following checks indicate FailureReason is related to a failed update
+		// Note: Internal service error is an exception for this case
+		// "Request to service failed" means update failed because of ISE and can be retried
+		(latestStatus != nil && failureReason != nil && lastEndpointConfigForUpdate != nil &&
+			!strings.HasPrefix(*failureReason, FailureReasonInternalServiceErrorPrefix) &&
+			*desiredEndpointConfig != *latestEndpointConfig &&
```
jkuruba, line 75: *desiredEndpointConfig != *latestEndpointConfig is not necessary since you are checking for *desiredEndpointConfig == *lastEndpointConfigForUpdate

pkg/resource/endpoint/custom_update_api.go
```diff
@@ -19,21 +19,65 @@ package endpoint
...
-	diffReporter *ackcompare.Reporter,
+	delta *ackcompare.Delta,
 ) (*resource, error) {
-	return nil, rm.endpointStatusAllowUpdates(ctx, latest)
+	latestStatus := latest.ko.Status.EndpointStatus
+	if latestStatus == nil {
+		return nil, nil
+	}
+
+	if latestStatus != nil && *latestStatus != svcsdk.EndpointStatusFailed {
```
RedbackThomson, line 54: This issue has not been resolved?

pkg/resource/endpoint/custom_update_conditions.go
RedbackThomson: The terminal condition may be set, but set to False (that case is not covered on lines 51-54)?

pkg/resource/endpoint/custom_update_api.go
surajkota: good question! both are needed

1. FailureReason alone does mean an update failed it can appear because of other reasons(patching/scaling failed) 
2. `*desiredEndpointConfig == *lastEndpointConfigForUpdate` only tells us an update was tried with `lastEndpointConfigForUpdate` but does not tell us anything if the update was successful or not in the past because it is set if updateEndpoint returns 200 (remember aync operation). 
3. Now, sdkUpdate executes because of change in any field in Spec (like tags/deploymentConfig in future)

1&2 does not guarantee an update Failed. Hence we need to look at `*latestEndpointConfigName` to determine if the update was unsuccessful
`*desiredEndpointConfig != *latestEndpointConfig` + `*desiredEndpointConfig == *lastEndpointConfigForUpdate`+ `FailureReason != nil` indicate that an update is needed, has already been tried and failed.

That being said. I will add this information in the comments because I will forget it as well

pkg/resource/endpoint/custom_update_api.go
surajkota: just pushed

surajkota:
> The terminal condition may be set, but set to False (that case is not covered on lines 51-54)?
yep
https://github.com/surajkota/ack-sagemaker-controller/blob/endpoint_failed_update/pkg/resource/endpoint/custom_update_conditions.go#L65

RedbackThomson (approved):

mbaijal:
>      * This PR fixes the above problem by recording the endpointconfigName used for update under `Status.LastEndpointConfigNameForUpdate` and only retry an update if the `FailureReason` was related to an internal server error or if the user has submitted a new endpointConfigName. Additionally, it sets a `ACK.Terminal `condition to `True` and stops reconciling if either of these conditions is not met. User is expected to check the `failureReason` and re-submit the resource with a **new** endpointConfigName if they wish to update the endpoint.
 
Will this conditional check require any additional attention if we start using autogenerated names (I don't think so, but bringing it up so we don't miss it.)

surajkota:
> Will this conditional check require any additional attention if we start using autogenerated names (I don't think so, but bringing it up so we don't miss it.)

I don't think so because those would be generated before creation

generator.yaml
```diff
@@ -15,6 +15,8 @@ operations:
   UpdateEndpoint:
     custom_implementation: customUpdateEndpoint
     set_output_custom_method_name: customUpdateEndpointSetOutput
+    override_values:
+      RetainAllVariantProperties: true
```
mbaijal, line 19: This means that the user cannot customize this value correct  ?

pkg/resource/endpoint/custom_set_output.go
```diff
@@ -47,6 +47,19 @@ func (rm *resourceManager) customDescribeEndpointSetOutput(
 	ko *svcapitypes.Endpoint,
 ) (*svcapitypes.Endpoint, error) {
 	rm.customSetOutput(r, resp.EndpointStatus, ko)
+	// Workaround: Field config for LatestEndpointConfigName of generator config
```
mbaijal, line 50: major nit: Add punctuation to comments, took me a couple of rereads to realize these were different sentences.

pkg/resource/endpoint/custom_set_output.go
```diff
@@ -47,6 +47,19 @@ func (rm *resourceManager) customDescribeEndpointSetOutput(
...
+	// does not code generate this correctly since this field is part of Spec
+	// SageMaker users will need following information:
```
mbaijal, line 52: users or the controller ?

mbaijal (approved):
Overall looks ok since I have been working with this code and it works as expected.

generator.yaml
surajkota: Correct

pkg/resource/endpoint/custom_set_output.go
surajkota: will do in a future PR

pkg/resource/endpoint/custom_set_output.go
surajkota: users

PR #8, opened 2021-03-23
Add Model Monitor Custom Resources

RedbackThomson:
Updates the SageMaker controller to the latest runtime (0.0.5)

Adds the following custom resources:
- `DataQualityJobDefinition`
- `ModelQualityJobDefinition`
- `ModelBiasJobDefinition`
- `ModelExplainabilityJobDefinition`
- `MonitoringSchedule`

All job definitions act as database entries (do not require reconciliation and do not support updates)
`MonitoringSchedule` has two unique status fields (`FailureReason` and `MonitoringScheduleStatus`)


generator.yaml
```diff
@@ -132,6 +132,49 @@ resources:
...
+            message_prefix: Monitoring Job Definition
+  MonitoringSchedule:
+    fields:
+      MonitoringScheduleStatus:
+        is_read_only: true
+        is_printable: true
+        from:
+          operation: DescribeMonitoringSchedule
+          path: MonitoringScheduleStatus
+      FailureReason:
```
surajkota, line 167: Why would a schedule be in a Pending or Failed Status and how would customer recover from failures

We need to treat this as a job and not like config

surajkota (changes requested):

generator.yaml
RedbackThomson: I disagree. The only commonality between this and a job is that it requires constant reconciliation to update the status fields. Otherwise it behaves exactly like a configuration. `Pending` or `Failed` statuses indicate that they have an error in their job definition, which caused the processing job running the model monitoring task to fail. For a user this would require updating their job definition - it has no effect on the scheduling. These status fields are conveniences to understand the previous state of your model monitoring invocations, but not the status of your schedule itself.

surajkota: > Pending or Failed status indicate that they have an error in their job definition, which caused the processing job running the model monitoring task to fail

I think this is not correct. the schedule itself can fail to register or update independent of the processing job.
FailureReason of the monitoring schedule is independent of the processing job.

If I understand if correctly, create_monitoring_schedule will return success and put the schedule to `Pending` status and user need to do a `describe_` until it is in `Scheduled`

and user also has the ability to update the monitoring schedule in which case there will be a similar lifecycle

jaypipes (changes requested):
@RedbackThomson let's do a different PR that regenerates the existing SageMaker controller for the v0.0.5 ACK runtime please. That way, we can have this PR focus only on the new resources.

surajkota:
Closing this PR to split it across Job definitions and Monitoring schedule with update generator.yaml after [aws-controllers-k8s/community#737](https://github.com/aws-controllers-k8s/community/issues/737) is closed

PR #11, opened 2021-04-09
[do-not-merge] SageMaker v0.0.1-beta Helm charts

mbaijal:
These are beta helm charts for local testing only. 
The Docker Image is pushed to `mbaijal/ack`.

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -1,0 +1,18 @@
+apiVersion: v1
+name: ack-sagemaker-controller
+description: A Helm chart for the ACK service controller for sagemaker
+version: v0.0.1
+appVersion: v0.0.1
+home: https://github.com/aws-controllers-k8s/sagemaker-controller
+icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
+sources:
+  - https://github.com/aws-controllers-k8s/sagemaker-controller
+maintainers:
+  - name: ACK Admins
+    url: https://github.com/orgs/aws-controllers-k8s/teams/ack-admin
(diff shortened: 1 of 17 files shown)


helm/Chart.yaml
```diff
@@ -0,0 +1,18 @@
...
+appVersion: v0.0.1
+home: https://github.com/aws-controllers-k8s/sagemaker-controller
+icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
+sources:
+  - https://github.com/aws-controllers-k8s/sagemaker-controller
+maintainers:
+  - name: ACK Admins
+    url: https://github.com/orgs/aws-controllers-k8s/teams/ack-admin
+  - name: sagemaker Admins
+    url:https://github.com/orgs/aws-controllers-k8s/teams/sagemaker-maintainer
```
RedbackThomson, line 14: hmm seems we missed a space in this field

helm/values.yaml
```diff
@@ -0,0 +1,43 @@
+# Default values for ack-sagemaker-controller.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+image:
+  repository: mbaijal/ack
```
RedbackThomson, line 6: I'm not sure if this is a good idea that we should push it into the official Github repository. One alternative is that you provide a values override file. As part of canaries or tests you can download and access these helm charts and then submit, when applying the chart to your cluster, with another values file overriding this value.
See https://stackoverflow.com/questions/51097553/can-i-have-multiple-values-yaml-files-for-helm/51099510 for information about applying an alternate values file when installing a chart.

PR #14, opened 2021-04-16
Add Model Monitor Job Definitions

surajkota:
Adds the following custom resources related to Model monitoring:

- DataQualityJobDefinition
- ModelQualityJobDefinition
- ModelBiasJobDefinition
- ModelExplainabilityJobDefinition

All job definitions act as database entries (do not require re-queuing and do not support updates)

PR includes Integration tests for the above resources and few minor enhancements:
 - reduce code duplication related to sagemaker_client
 - assert if deletion was successful. Haven't made this change to all the tests because another refractor PR is being worked on by @mbaijal and it will create a lot of merge conflicts
 - add asserts for dependent resources in endpoint and endpointconfig test
 - add a wait loop for endpoint deletion from sagemaker side
 - `go mod tidy`

### Testing
Locally using kind cluster

Formated test files using [black](https://black.readthedocs.io/en/stable/)

Related PRs:
[aws-controllers-k8s/code-generator#44](https://github.com/aws-controllers-k8s/code-generator/pull/44)


test/e2e/requirements.txt
```diff
@@ -1,4 +1,4 @@
-acktest @ git+https://github.com/aws-controllers-k8s/test-infra.git@6c6d9dd4759c7bb2feecd11dc03e8516371812cc
+acktest @ git+https://github.com/surajkota/ack-test-infra@delete_helper_function
```
surajkota, line 1: waiting for test-infra PR to be merged

test/e2e/common/fixtures.py
```diff
@@ -0,0 +1,110 @@
...
+    assert endpoint_resource is not None
+    assert k8s.get_resource_arn(endpoint_resource) is not None
+    wait_sagemaker_endpoint_status(
+        sagemaker_client, replacements["ENDPOINT_NAME"], "InService"
+    )
+
+    yield endpoint_spec
+
+    for cr in (model_reference, config_reference, endpoint_reference):
+        assert k8s.assert_delete_resource_successful(cr)
```
jaypipes, line 83: I don't think the PR in test-infra that adds the `assert_delete_resource_successful` function will be merged. Not only does the `assert_delete_resource_successful` function not actually assert anything, but it's actually more readable to just write:

```python
_, deleted = k8s.delete_custom_resource(cr)
assert(deleted)
```

alternately, I'd be willing to entertain updating the `k8s.delete_custom_resource` function to just return `True` or `False` instead of a tuple of response object and boolean.

test/e2e/requirements.txt
```diff
@@ -32,3 +32,4 @@ six==1.15.0
 toml==0.10.2
 urllib3==1.26.4
 websocket-client==0.58.0
+black==20.*
```
jaypipes, line 35: black isn't a requirement for the test code itself. It's a tool used in testing.

jaypipes (changes requested):

test/e2e/requirements.txt
surajkota: yes, for linting, going forward all testing PRs in this repo will be formatted using this tool and in future, we plan to add it as a check in PR builds

test/e2e/common/fixtures.py
```diff
@@ -0,0 +1,112 @@
...
+
+    yield endpoint_spec
+
+    for cr in (model_reference, config_reference, endpoint_reference):
+        _, deleted = k8s.delete_custom_resource(cr)
+        assert deleted
+
+
+@pytest.fixture(scope="module")
+def xgboost_churn_data_quality_job_definition(xgboost_churn_endpoint):
```
surajkota, line 88: FYI: this fixture is here since I plan to use it for monitoring schedule test later

test/e2e/__init__.py
```diff
@@ -51,3 +54,25 @@ def create_sagemaker_resource(
     )
 
     return reference, spec, resource
+
+
+def wait_sagemaker_endpoint_status(
```
akartsky, line 59: this should return `true` or `false` depending on whether it passed or not 

you can also return `actual_status` along with the Boolean

test/e2e/common/fixtures.py
```diff
@@ -0,0 +1,112 @@
...
+    """Creates a SageMaker endpoint with the XGBoost churn single-variant model
+    and data capture enabled.
+    """
+    endpoint_resource_name = random_suffix_name("xgboost-churn", 32)
+    config_resource_name = endpoint_resource_name + "-config"
+    model_resource_name = config_resource_name + "-model"
+
+    replacements = REPLACEMENT_VALUES.copy()
+    replacements["ENDPOINT_NAME"] = endpoint_resource_name
+    replacements["CONFIG_NAME"] = config_resource_name
```
akartsky, line 43: can we say `ENDPOINT_CONFIG_NAME` instead of `CONFIG_NAME`

jkuruba (approved):

test/e2e/tests/test_data_quality_job_definition.py
```diff
@@ -0,0 +1,67 @@
...
+        (reference, resource) = xgboost_churn_data_quality_job_definition
+        assert k8s.get_resource_exists(reference)
+
+        job_definition_name = resource["spec"].get("jobDefinitionName")
+        assert (
+            k8s.get_resource_arn(resource)
+            == describe_sagemaker_data_quality_job_definition(
+                sagemaker_client, job_definition_name
+            )["JobDefinitionArn"]
+        )
```
akartsky, line 57: i think we should consider changing `black` line-length to `110` instead of it's default `88`

test/e2e/tests/test_endpoint.py
```diff
@@ -56,10 +52,12 @@ def single_container_model(name_suffix):
         replacements=replacements,
     )
     assert model_resource is not None
+    assert k8s.get_resource_arn(model_resource) is not None
 
     yield (model_reference, model_resource)
 
-    k8s.delete_custom_resource(model_reference)
+    _, deleted = k8s.delete_custom_resource(model_reference)
+    assert deleted
```
akartsky, line 60: nice

test/e2e/tests/test_endpoint.py
```diff
@@ -410,16 +391,21 @@ def delete_endpoint_test(self, sagemaker_client, xgboost_endpoint):
...
-        # If we see errors in this part of test, can add a loop in future or consider changing controller
-        # to wait for SageMaker
-        time.sleep(10)
-        assert (
-            self._describe_sagemaker_endpoint(sagemaker_client, endpoint_name) is None
-        )
+        # We wait maximum of 30 seconds for this clean up to happen
+        endpoint_deleted = False
+        for _ in range(6):
+            time.sleep(5)
```
akartsky, line 401: i think its more reasonable to do `range(3)` and `sleep(10)`

generator.yaml
```diff
@@ -290,6 +290,90 @@ resources:
         from:
           operation: DescribeHyperParameterTuningJob
           path: OverallBestTrainingJob
+  DataQualityJobDefinition:
+    exceptions:
+      errors:
+          404:
+            code: ResourceNotFound
+      terminal_codes:
+        - ResourceLimitExceeded
```
mbaijal, line 299: is there any ambiguity here on which error codes we add here - as in does it vary per resource ? I was away so missed most of these conversations, if any.

go.mod
```diff
@@ -6,7 +6,6 @@ require (
 	github.com/aws-controllers-k8s/runtime v0.0.6
 	github.com/aws/aws-sdk-go v1.38.11
 	github.com/go-logr/logr v0.1.0
-	github.com/google/go-cmp v0.3.1
```
mbaijal, line 9: hm ? upstream change ?

test/e2e/__init__.py
```diff
@@ -24,6 +26,7 @@
 CONFIG_RESOURCE_PLURAL = "endpointconfigs"
 MODEL_RESOURCE_PLURAL = "models"
 ENDPOINT_RESOURCE_PLURAL = "endpoints"
+DATA_QUALITY_JOB_DEFINITION_RESOURCE_PLURAL = "dataqualityjobdefinitions"
```
mbaijal, line 29: why are these defined here vs resource test files ?
hpo/processing are still doing it differently. We will need to review test PRs carefully for a while to catch these diffs.

test/e2e/common/fixtures.py
```diff
@@ -0,0 +1,112 @@
...
+    wait_sagemaker_endpoint_status,
+)
+
+from e2e.replacement_values import REPLACEMENT_VALUES
+from acktest.resources import random_suffix_name
+from acktest.k8s import resource as k8s
+
+
+@pytest.fixture(scope="module")
+def xgboost_churn_endpoint(sagemaker_client):
```
mbaijal, line 33: Could you separate out the model and endpoint creation  such that we can use model other places such as the transform job ?

test/e2e/requirements.txt
mbaijal: How about adding the pytest-black plugin to check for zero diffs as an integ test already ? or is there some reason we dont want this in our integration tests ?

test/e2e/tests/test_endpoint.py
mbaijal: fair point, could reduce the checks by half with an at max lag extra lag of 5seconds.

test/e2e/__init__.py
```diff
@@ -51,3 +54,25 @@ def create_sagemaker_resource(
...
+def wait_sagemaker_endpoint_status(
+    sagemaker_client,
+    endpoint_name,
+    expected_status: str,
+    wait_periods: int = 60,
+):
+    actual_status = None
+    for _ in range(wait_periods):
+        time.sleep(30)
+        actual_status = sagemaker_client.describe_endpoint(EndpointName=endpoint_name)[
```
mbaijal, line 68: If we are moving this out to a common file, we should make this common across resources. One way could be to pass  `sagemaker_client.describe_endpoint(EndpointName=endpoint_name)` as a function parameter to the `wait_sagemaker_generic_status` function.

test/e2e/tests/test_model_bias_job_definition.py
```diff
@@ -0,0 +1,94 @@
...
+    except BaseException:
+        logging.error(
+            f"Could not find Model Bias Job Definition with name {job_definition_name}"
+        )
+        return None
+
+
+@service_marker
+@pytest.mark.canary
+class TestModelBiasJobDefinition:
```
mbaijal, line 73: Since I don't understand these resources completely, just checking that none of these require requeues and tests to completion, correct ?

test/e2e/tests/test_trainingjob.py
```diff
@@ -29,11 +29,6 @@
 RESOURCE_PLURAL = "trainingjobs"
 
 
-@pytest.fixture(scope="module")
-def sagemaker_client():
```
mbaijal, line 33: Thank you!

test/e2e/tests/test_hpo.py
```diff
@@ -86,5 +82,5 @@ def test_create_hpo(self, xgboost_hpojob):
         _, deleted = k8s.delete_custom_resource(reference)
         assert deleted is True
 
-        hpo_sm_desc = get_sagemaker_hpo_job(hpo_job_name)
+        hpo_sm_desc = get_sagemaker_hpo_job(sagemaker_client, hpo_job_name)
```
akartsky, line 85: we cannot keep using the same boto3 client object `sagemaker_client`

the session in the object might expire 

when the credentials are refreshed the new credentials won't be picked up if we don't create new session (or new client like we used to before you made this change)

go.mod
jaypipes: Yeah, this is due to runtime v0.0.6 "delta" work replacing the use of the above library.

jaypipes (approved):

test/e2e/tests/test_endpoint.py
surajkota: done

test/e2e/tests/test_model_bias_job_definition.py
surajkota: Correct, only support Create, delete

test/e2e/tests/test_hpo.py
surajkota: as discussed offline, good suggestion but needs to be in a separate PR
hpo was an doing it differently
this PR is just moving the fixture to a different file not changing the existing behaviour

test/e2e/__init__.py
surajkota: lets change it when needed

go.mod
surajkota: + `go mod tidy`

test/e2e/__init__.py
surajkota: resources that are being used by multiple tests have plurals defined here in a common location

test/e2e/common/fixtures.py
surajkota: we already have a helper method create_sagemaker_resource which is generic, everything else is specific to a use case because the replacements and spec file names are hardcoded.

If you find a transform job example in future that uses this xgboost_churn model, we can move it out easily

will keep as-is for now

test/e2e/tests/test_data_quality_job_definition.py
surajkota: agree with this but let me open a follow up PR for this because running with 110 seems to be touching all the files under test dir

test/e2e/requirements.txt
surajkota: couldn't find a way to run this tool as part of integ test, it looks like a CLI tool

so we can use it when we have PR builds to check if it returns any diff

test/e2e/__init__.py
mbaijal: I think we should make the change now, since I was reusing your logic from endpoint status wait across all other tests. That is a lot of duplication and this is how I was planning to fix it if I got the chance.

test/e2e/common/fixtures.py
mbaijal: I am using almost similar code to create the transform job model object. If we parameterized the s3 path, code could be re-used right away.

akartsky (approved):

test/e2e/__init__.py
surajkota: got it, made it generic

test/e2e/common/fixtures.py
mbaijal: resolving this, I will make this change as a part of my PR if it actually dedups some logic.

mbaijal (approved):

PR #17, opened 2021-04-22
Add Model Monitor - Monitoring Schedule

surajkota:
### Description of changes
- Add Monitoring schedule
- Integration tests
- Custom code to handle specialized logic related to update, delete and setting terminal status for Failed status. I have added comments for each custom code which should provide clarity for review

### Testing
- e2e lifecycle using automated tests locally using pytest
- Manually tested for on going baselining job


pkg/resource/monitoring_schedule/custom_update_api.go
```diff
@@ -0,0 +1,50 @@
...
+	if r.ko.Status.LastMonitoringExecutionSummary != nil {
+		executionStatus := r.ko.Status.LastMonitoringExecutionSummary.MonitoringExecutionStatus
+		if *executionStatus == "Pending" || *executionStatus == "InProgress" || *executionStatus == "Stopping" {
+			inProgressExecutions = true
+		}
+	}
+	if scheduleStatus != nil && (*scheduleStatus == "Pending" || inProgressExecutions) {
+		return requeue.NeededAfter(
+			errors.New("monitoring schedule status does not allow modification, it is in-progress"),
+			requeue.DefaultRequeueAfterDuration)
```
RedbackThomson, line 46: This is going to look transparent to the users? They will submit a new spec and then the reconciler will do nothing until it reaches the correct backend state? I wonder if we can flow this back up to the users.

pkg/resource/monitoring_schedule/custom_update_api.go
```diff
@@ -0,0 +1,50 @@
...
+	scheduleStatus := r.ko.Status.MonitoringScheduleStatus
+	inProgressExecutions := false
+	// It is possible that schedule doesn't have any execution yet
+	if r.ko.Status.LastMonitoringExecutionSummary != nil {
+		executionStatus := r.ko.Status.LastMonitoringExecutionSummary.MonitoringExecutionStatus
+		if *executionStatus == "Pending" || *executionStatus == "InProgress" || *executionStatus == "Stopping" {
+			inProgressExecutions = true
+		}
+	}
+	if scheduleStatus != nil && (*scheduleStatus == "Pending" || inProgressExecutions) {
```
RedbackThomson, line 43: I think we can clean up these conditionals by doing an early exit on line 36:
```
if r.ko.Status.LastMonitoringExecutionSummary == nil || scheduleStatus == nil {
  return nil
}
```

test/e2e/tests/test_monitoring_schedule.py
```diff
@@ -0,0 +1,219 @@
...
+def xgboost_churn_data_quality_monitoring_schedule(
+    xgboost_churn_data_quality_job_definition,
+):
+    (_, job_definition_resource) = xgboost_churn_data_quality_job_definition
+
+    job_definition_name = job_definition_resource["spec"].get("jobDefinitionName")
+
+    monitoring_schedule_name = random_suffix_name("monitoring-schedule", 32)
+
+    replacements = REPLACEMENT_VALUES.copy()
```
RedbackThomson, line 45: Hmm it would be cool to just pass the replacement values and it merges it into `REPLACEMENT_VALUES` inside `create_sagemaker_resource`.

test/e2e/tests/test_monitoring_schedule.py
```diff
@@ -0,0 +1,219 @@
...
+        # Pending status is shortlived only for 30 seconds because baselining job has already been run
+        # remove the checks for Pending status if the test is flaky because of this
+        # as the main objective is to test for Scheduled status
+        # OR
+        # create the schedule with a on-going baseline job where it waits for the baselining job to complete
+        assert (
+            wait_resource_monitoring_schedule_status(
+                reference, self.status_pending, 5, 2
+            )
+            == self.status_pending
```
RedbackThomson, line 170: Are we certain it will be in pending? It could skip over pending if we don't poll it immediately?

test/e2e/tests/test_monitoring_schedule.py
```diff
@@ -0,0 +1,219 @@
...
+        get_monitoring_schedule_resource_status,
+        reference,
+    )
+
+
+@service_marker
+@pytest.mark.canary
+class TestMonitoringSchedule:
+    status_pending: str = "Pending"
+    status_scheduled: str = "Scheduled"
```
RedbackThomson, line 128: nit: These constants should be all uppercase `STATUS_PENDING`

test/e2e/tests/test_monitoring_schedule.py
```diff
@@ -0,0 +1,219 @@
...
+        assert k8s.wait_on_condition(reference, "ACK.ResourceSynced", "True")
+
+        # Update the resource
+        new_cron_expression = "cron(0 * * * ? *)"
+        spec["spec"]["monitoringScheduleConfig"]["scheduleConfig"][
+            "scheduleExpression"
+        ] = new_cron_expression
+        resource = k8s.patch_custom_resource(reference, spec)
+        resource = k8s.wait_resource_consumed_by_controller(reference)
+        assert resource is not None
```
RedbackThomson, line 186: Considering the custom update logic, let's just put this into a separate test? That way you can have finer control over the tests for update rather than lumping it into smoke

test/e2e/tests/test_monitoring_schedule.py
```diff
@@ -0,0 +1,219 @@
...
+        )
+
+        # Delete the k8s resource.
+        _, deleted = k8s.delete_custom_resource(reference)
+        assert deleted
+
+        # 30 sec wait for server-side acknowledgement
+        schedule_deleted = False
+        for _ in range(3):
+            time.sleep(10)
```
RedbackThomson, line 210: 30 seconds for acknowledgement? That's a long time between API call and response

pkg/resource/monitoring_schedule/custom_update_api.go
surajkota: It does via `ACK.Recoverable` condition

pkg/resource/monitoring_schedule/custom_update_api.go
surajkota: added only
```
if scheduleStatus == nil {
		return nil
	}
```

since is possible that schedule doesn't have any execution yet

test/e2e/tests/test_monitoring_schedule.py
surajkota: edited the comment
its wait for server-side clean up

test/e2e/tests/test_monitoring_schedule.py
surajkota: all resources now have some custom code which make it deviate from default generated code so I think this is ok and anyway these operations need to happen in order

test/e2e/tests/test_monitoring_schedule.py
surajkota: yes and these checks need to happen in 30 seconds
will see and iterate on this

test/e2e/tests/test_monitoring_schedule.py
surajkota: will try this out in a separate PR

RedbackThomson (approved):

mbaijal (approved):

PR #19, opened 2021-04-26
Job-style resource testing improvements

mbaijal:
#### DESCRIPTION OF CHANGES: 
1. Adds a test to completion for all job type resources. 
2. Refactored tests for processing, training, debugger, batchtransform to match hpo. Another iteration needed here to bring all test files in sync. 
3. Fixed error in processingJob test yaml (typo in bucket variable name)
4. Fixed error in the batch transform's model file yaml which used an incorrect xgboost image version - this should not affect existing tests in any way (tests are passing)
5.  Added test for trainingjob with debugger along with necessary replacement values and tests to completion.

#### TESTING:
Local Tests passed on previous the commit (HEAD -1) in progress on the latest rebase.


test/e2e/tests/test_hpo.py
```diff
@@ -18,40 +18,56 @@
...
 
-    reference, spec, resource = create_sagemaker_resource(
-        resource_plural=RESOURCE_PLURAL,
-        resource_name=resource_name,
-        spec_file="xgboost_hpojob",
-        replacements=replacements,
+    data = load_resource_file(
+        resource_directory, "xgboost_hpojob", additional_replacements=replacements
+    )
+    logging.debug(data)
```
jaypipes, line 49: did you mean to leave this in there?

test/e2e/tests/test_hpo.py
```diff
@@ -18,40 +18,56 @@
...
+    create_sagemaker_resource,
+    wait_for_status,
+)
+from e2e.replacement_values import REPLACEMENT_VALUES
+from e2e.bootstrap_resources import get_bootstrap_resources
+from time import sleep
 
 RESOURCE_PLURAL = "hyperparametertuningjobs"
-HPO_JOB_STATUS_CREATED = ("InProgress", "Completed")
-HPO_JOB_STATUS_STOPPED = ("Stopped", "Stopping")
```
jaypipes, line 28: Any reason you moved these out of module scope and into class scope of the `TestHPO` class?

test/e2e/tests/test_hpo.py
```diff
@@ -61,26 +77,99 @@ def get_sagemaker_hpo_job(sagemaker_client, hpo_job_name):
...
+        )
+
+        # Delete the k8s resource.
+        _, deleted = k8s.delete_custom_resource(reference)
+        assert deleted is True
+
+        hpo_sm_desc = get_sagemaker_hpo_job(hpo_job_name)
+        assert hpo_sm_desc["HyperParameterTuningJobStatus"] in self.list_status_stopped
+
+    def test_completed_hpo(self, xgboost_hpojob):
```
jaypipes, line 154: Any reason you are creating another test method (and another HPO Job) instead of keeping the checks on status in the original test method that tests the full "happy path" lifecycle of an HPO Job?

jaypipes (commented):
I actually think these changes make the tests *harder* to read and not easier... Also, I'm not sure why we are creating multiple resources just for asserting that a resource gets into some Completing status when the original single test methods were already checking this status field transition.

test/e2e/tests/test_hpo.py
mbaijal: Seemed cleaner to me this way but on second thought, I could actually move all of these out to a shared config since the values of these statuses are the same across job style resources.

test/e2e/tests/test_hpo.py
mbaijal: Yes, this is essential. 
1. The first test starts the hpoJob -> checks that it gets created -> then stops -> checks that the status in is stopping/stopped in SM and the resource gets deleted from k8s. 
2. The second (new) test starts the hpoJob -> checks that it gets created -> then waits for it to go to completion in SageMaker making sure all the fields work as expected -> it also checks that the k8s status gets updated as expected, ensuring requeues (testing custom code) -> and finally tests that a job completed in SageMaker, can be deleted from k8s (a bug we caught recently).

This second case has already caught a couple of bugs - although just incorrect specs - but then those specs will become our samples so we do need to test both scenarios.

test/e2e/tests/test_hpo.py
mbaijal: No, fixed it.

test/e2e/resources/kmeans_processingjob.yaml
```diff
@@ -47,6 +47,6 @@ spec:
           s3UploadMode: EndOfJob
       - outputName: valid_data
         s3Output:
-          s3URI: s3://{DATA_BUCKET}/sagemaker/processing/output/
+          s3URI: s3://$SAGEMAKER_DATA_BUCKET/sagemaker/processing/output/
```
surajkota, line 50: lol!

test/e2e/tests/test_hpo.py
surajkota: also sdkDelete in 
1. calls SageMaker Stop*Job() 
2. retuns early without calling Stop*Job() - again testing custom code
having separate tests ensure both cases work

test/e2e/tests/test_hpo.py
```diff
@@ -18,40 +18,55 @@
...
 
-    # Delete the k8s resource if not already deleted by tests
-    if k8s.get_resource_exists(reference):
-        k8s.delete_custom_resource(reference)
+    return reference, data
+
+@pytest.fixture(scope="function")
+def xgboost_hpojob():
+    hpo_job, data = _make_hpojob()
+    resource = k8s.create_custom_resource(hpo_job, data)
```
surajkota, line 59: in all tests,

1. use `create_sagemaker_resource` to reduce amount of code. basically, this test doesnt need make_hpojob at all

2. and instead of 
  ```
  if k8s.get_resource_exists(hpo_job):
          k8s.delete_custom_resource(hpo_job)
  ```
  do
  ```
      _, deleted = k8s.delete_custom_resource(reference)
      assert deleted
  ```

test/e2e/tests/test_hpo.py
surajkota: +1 to move to common file

test/e2e/tests/test_hpo.py
```diff
@@ -61,26 +76,102 @@ def get_sagemaker_hpo_job(sagemaker_client, hpo_job_name):
...
+    def test_hpo(self, xgboost_hpojob):
+        (reference, resource) = xgboost_hpojob
+        assert k8s.get_resource_exists(reference)
+    
+        hpo_job_name = resource["spec"].get("hyperParameterTuningJobName", None)
+        assert hpo_job_name is not None
+
+        hpo_sm_desc = get_sagemaker_hpo_job(hpo_job_name)
+        assert k8s.get_resource_arn(resource) == hpo_sm_desc["HyperParameterTuningJobArn"]
+        assert hpo_sm_desc["HyperParameterTuningJobStatus"] in self.list_status_created
```
surajkota, line 142: for all tests, I think we should remove `Completed` from `list_status_created`  because immidiately after this we are checking for `"ACK.ResourceSynced", "False"` which is contradictory

`Completed` should actually be part of `list_status_stopped` because job can actually go from `Stopping` -> `Completed` if it was stopped when it was almost finished(like model was being uploaded)

test/e2e/tests/test_hpo.py
```diff
@@ -18,40 +18,55 @@
...
+)
+from e2e.replacement_values import REPLACEMENT_VALUES
+from e2e.bootstrap_resources import get_bootstrap_resources
+from time import sleep
 
 RESOURCE_PLURAL = "hyperparametertuningjobs"
-HPO_JOB_STATUS_CREATED = ("InProgress", "Completed")
-HPO_JOB_STATUS_STOPPED = ("Stopped", "Stopping")
 
+def _sagemaker_client():
```
surajkota, line 37: dont need this in test files, defined as a fixture in conftest.py in one of the earlier PRs

test/e2e/tests/test_hpo.py
```diff
@@ -61,26 +76,102 @@ def get_sagemaker_hpo_job(sagemaker_client, hpo_job_name):
...
         hpo_job_name = resource["spec"].get("hyperParameterTuningJobName", None)
         assert hpo_job_name is not None
 
-        hpo_sm_desc = get_sagemaker_hpo_job(sagemaker_client, hpo_job_name)
+        hpo_sm_desc = get_sagemaker_hpo_job(hpo_job_name)
         assert (
             k8s.get_resource_arn(resource) == hpo_sm_desc["HyperParameterTuningJobArn"]
         )
-        assert hpo_sm_desc["HyperParameterTuningJobStatus"] in HPO_JOB_STATUS_CREATED
+        assert hpo_sm_desc["HyperParameterTuningJobStatus"] in self.list_status_created
```
surajkota, line 167: _assert_hpo_status_in_sync ?
nit: remove _ prefix from method names

test/e2e/tests/test_hpo.py
surajkota: lets just call it `test_completed` and `test_stopped` ? because the its already under TestHPO class

test/e2e/tests/test_trainingjob_debugger.py
```diff
@@ -0,0 +1,243 @@
...
+        )
+
+        # Delete the k8s resource.
+        _, deleted = k8s.delete_custom_resource(reference)
+        assert deleted is True
+
+        training_job_desc = get_sagemaker_training_job(training_job_name)
+        assert training_job_desc["TrainingJobStatus"] in self.list_status_stopped
+
+    def test_completed_training_job_debugger(self, xgboost_training_job_debugger):
```
surajkota, line 217: should we just make this the test to completion for training job? do we really need a separate test?

IMO, even if we do`test_training_job_debugger` is redundant and remove canary_marker

surajkota (changes requested):

test/e2e/tests/test_hpo.py
jaypipes: @mbaijal @surajkota why are we deleting the Kubernetes CR when a job "completes"? I thought we only wanted to delete jobs when a user explicitly deleted the job CR in Kubernetes?

surajkota: I think there is some miscommunication, we are testing the case where the user deletes the CR after the job is complete, controller does not delete the CR automatically

Sounds good ?

test/e2e/replacement_values.py
```diff
@@ -42,6 +42,54 @@
     "sa-east-1": "737474898029.dkr.ecr.sa-east-1.amazonaws.com",
 }
 
+XGBOOST_DEBUGGER_IMAGE_URIS = {
```
surajkota, line 45: nit: ~~XGBOOST_~~ DEBUGGER

test/e2e/replacement_values.py
```diff
@@ -42,6 +42,54 @@
...
+    "eu-north-1": 	    "314864569078.dkr.ecr.eu-north-1.amazonaws.com",
+    "eu-west-1": 	    "929884845733.dkr.ecr.eu-west-1.amazonaws.com",
+    "eu-west-2": 	    "250201462417.dkr.ecr.eu-west-2.amazonaws.com",
+    "eu-west-3": 	    "447278800020.dkr.ecr.eu-west-3.amazonaws.com",
+    "me-south-1": 	    "986000313247.dkr.ecr.me-south-1.amazonaws.com",
+    "sa-east-1": 	    "818342061345.dkr.ecr.sa-east-1.amazonaws.com"
+}
+
+# https://docs.aws.amazon.com/sagemaker/latest/dg/sagemaker-algo-docker-registry-paths.html
+XGBOOST_TRANSFORM_MODEL_V1_IMAGE_URIS = {
```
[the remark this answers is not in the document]
mbaijal: Is there a reason we shouldn't use the API calls to get these image URIs, I was hoping to switch to that where possible.

test/e2e/tests/test_hpo.py
mbaijal: I started to make this change and did too across two resources but using this one as a fixture and passing it around every method in the class seems unnecessary. Is there are a strong (performance or otherwise) reason for doing that ? (considering it is not a ton of code duplication. 
Another option could be that we could define it not as a fixture but an imported method which would make the code more readable.

test/e2e/tests/test_hpo.py
mbaijal: Yea, this was one of the changes i said I saw after a more recent rebase. Done.

test/e2e/tests/test_hpo.py
mbaijal: I am following the naming across all files here, or did i miss something ?

test/e2e/tests/test_trainingjob_debugger.py
mbaijal: I will remove the canary marker from the training test, agreed. But since these are two different specs and a simple trainingjob is our most classic example, I do not want to remove it.

test/e2e/resources/kmeans_processingjob.yaml
mbaijal: Checkout the debugger Image Issue - it's worse :P

test/e2e/tests/test_trainingjob_debugger.py
```diff
@@ -0,0 +1,226 @@
...
+
+        assert k8s.get_resource_arn(resource) == training_job_desc["TrainingJobArn"]
+        assert training_job_desc["TrainingJobStatus"] == JOB_STATUS_INPROGRESS
+        assert k8s.wait_on_condition(reference, "ACK.ResourceSynced", "False")
+
+        self._assert_training_status_in_sync(
+            training_job_name, reference, JOB_STATUS_COMPLETED
+        )
+        # TODO: This test is failing
+        #assert k8s.wait_on_condition(reference, "ACK.ResourceSynced", "False")
```
surajkota, line 217: uncomment

test/e2e/tests/test_trainingjob_debugger.py
```diff
@@ -0,0 +1,226 @@
...
+    ):
+        assert (
+            self._wait_sagemaker_training_debugger_status(
+                training_job_name, expected_status
+            )
+            == self._wait_resource_training_debugger_status(reference, expected_status)
+            == expected_status
+        )
+    
+    def test_stopped(self, xgboost_training_job_debugger):
```
surajkota, line 176: do we need test stopped again, nothing different from TestTrainingJob::test_stopped right?

test/e2e/tests/test_transformjob.py
```diff
@@ -23,142 +23,165 @@
...
 )
 from e2e.replacement_values import REPLACEMENT_VALUES
 from e2e.bootstrap_resources import get_bootstrap_resources
+from e2e.common.config import *
+from time import sleep
 
-RESOURCE_PLURAL = "transformjobs"
+RESOURCE_PLURAL = 'transformjobs'
 
+def _make_model():
```
surajkota, line 35: - make this a fixture so you add delete in case the test fails
- if possible use the model name like `transform_resource_name+"-model"` so its easy to identify while debugging

test/e2e/tests/test_trainingjob_debugger.py
```diff
@@ -0,0 +1,226 @@
...
+            TrainingJobName=training_job_name
+        )
+        return training_job
+    except BaseException:
+        logging.error(
+            f"SageMaker could not find a training debugger job with the name {training_job_name}"
+        )
+        return None
+
+# TODO: Move to __init__.py
```
surajkota, line 71: nit: can we do it now?
not a blocker

test/e2e/tests/test_hpo.py
surajkota: you are right, sounds good, can we move this to a common file rather than in each test file?

test/e2e/tests/test_hpo.py
```diff
@@ -18,40 +18,47 @@
...
 from acktest.resources import random_suffix_name
 from acktest.k8s import resource as k8s
+from e2e import (
+    service_marker,
+    create_sagemaker_resource,
+    wait_for_status,
+)
+from e2e.replacement_values import REPLACEMENT_VALUES
+from e2e.bootstrap_resources import get_bootstrap_resources
+from e2e.common.config import *
```
surajkota, line 30: thanks!
it would be great if we can do `from e2e.common import config as cfg` instead of * because all are global variables

and then use variables like `cfg. JOB_STATUS_INPROGRESS`

surajkota (changes requested):

test/e2e/tests/test_hpo.py
surajkota: ya, this was comments in one of the previous PRs, doesn't make sense to prefix with _
ok for now, not a blocker

test/e2e/tests/test_trainingjob_debugger.py
mbaijal: fair point, but then if this is the only one running in the canary, we should keep it.

test/e2e/tests/test_trainingjob_debugger.py
mbaijal: keeping as TODO for now since as discussed this may not be straightforward.

test/e2e/tests/test_transformjob.py
mbaijal: I did the second for which I had to pass the model name as a parameter. 
Now when I make the make_model a fixture, I cannot use it as one since it needs a parameter. So I am using a fixture as a regular method, unless I want to make the transformjob name global.

test/e2e/common/config.py
```diff
@@ -0,0 +1,31 @@
...
+
+# Job Type Resource Statuses 
+LIST_JOB_STATUS_STOPPED = (
+    "Stopped", 
+    "Stopping", 
+    "Completed"
+    )
+JOB_STATUS_INPROGRESS: str = "InProgress"
+JOB_STATUS_COMPLETED: str = "Completed"
+DEBUGGERJOB_STATUS_COMPLETED: str = (
```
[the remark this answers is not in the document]
mbaijal: correct!

test/e2e/common/config.py
```diff
@@ -0,0 +1,31 @@
...
+"""String Constants
+"""
+
+ENDPOINT_CONFIG_RESOURCE_PLURAL = "endpointconfigs"
+MODEL_RESOURCE_PLURAL = "models"
+ENDPOINT_RESOURCE_PLURAL = "endpoints"
+DATA_QUALITY_JOB_DEFINITION_RESOURCE_PLURAL = "dataqualityjobdefinitions"
+
+# Job Type Resource Statuses 
+LIST_JOB_STATUS_STOPPED = (
```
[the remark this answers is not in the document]
mbaijal: I think thats ok, the variable name here indicates it is checking a list - tuple in the name sounds unintuitive.

test/e2e/tests/test_trainingjob_debugger.py
surajkota: I had a typo it should work

test/e2e/tests/test_transformjob.py
surajkota: Dont forget you clean up the model in transform fixture

test/e2e/tests/test_hpo.py
mbaijal: done

test/e2e/tests/test_trainingjob_debugger.py
mbaijal: I tried but still ran into issues using yield and parameterized fixtures. Plus I will need to change this across all tests. I'll create a separate PR but do not see value in spending too much time debugging this at the moment.

surajkota (approved):
Let's ship it!

test/e2e/tests/test_transformjob.py
```diff
@@ -23,142 +23,178 @@
...
+    reference, _, resource = create_sagemaker_resource(
         resource_plural=RESOURCE_PLURAL,
-        resource_name=resource_name,
+        resource_name=transform_resource_name,
         spec_file="xgboost_transformjob",
         replacements=replacements,
     )
+
     assert resource is not None
+    assert k8s.get_resource_arn(resource) is not None
```
surajkota, line 83: I think this check should be here because it means we are expecting sdkCreate to have succeeded. if it fails for some reason, the assert would fail and resource won't get cleaned up 
so, we should have a try-catch block in the fixture as well or better move this check to the test class

PR #20, opened 2021-04-28
Cleans the requirements.txt file

RedbackThomson:
Ensure the `requirements.txt` has as few components as possible.
If it has overlapping dependencies as the `acktest` package, it becomes difficult to update any security vulnerabilities. 

*All other dependencies are included as direct or indirect dependencies of `acktest`*


surajkota (approved):
Thanks for the PR! I was going to do this as well

PR #23, opened 2021-05-07
add canary related supporting infra

surajkota:
### Description
Scripts and build spec to run tests on existing EKS cluster using oidc and helm charts on codebuild

### Testing
Using codebuild

### Related PRs
https://github.com/aws-controllers-k8s/test-infra/pull/18


test/canary/scripts/run_test.sh
```diff
@@ -0,0 +1,81 @@
...
+      --sort-by=.metadata.creationTimestamp \
+      | grep ack-sagemaker-controller | awk '{print $1}' 2>/dev/null )
+
+  kubectl -n $NAMESPACE logs "$pod_id"
+}
+
+function cleanup {
+  echo "Cleaning up resources"
+  set +e
+  kubectl delete endpoints.sagemaker --all
```
akartsky, line 33: this means we won't be able to run codebuild tests in parallel as one test might delete resources being used in another test.

Right now we don't need this as we don't have a problem of leaking resources.
By doing this we will not come to know if there are some leaking resources in future. 

I think we should just handle these kinds of cleanup better in the the tests itself. 
(this will help while running tests locally too as they might leave resources behind)

test/e2e/tests/test_endpoint_config.py
```diff
@@ -26,7 +26,7 @@
     create_sagemaker_resource,
 )
 from e2e.replacement_values import REPLACEMENT_VALUES
-from e2e.common.config import config as cfg
+from e2e.common import config as cfg
```
akartsky, line 29: doesn't seem related to canaries ?

surajkota: yes, bug from a preview PR

test/canary/scripts/run_test.sh
akartsky: also there is the problem of finalizer where delete might get stuck forever

akartsky (approved):

PR #28, opened 2021-05-14
refactor tests and bug fixes

surajkota:
### Description of changes
- model and endpoint config tests missed in last refactoring PR
- fix wait_resource in adopt_endpoint test
- log boto3 errors
- increase delete timeout
- add owner aliases

### testing
using codebuild

### DIFF
--- a/test/e2e/tests/test_data_quality_job_definition.py
+++ b/test/e2e/tests/test_data_quality_job_definition.py
@@ -15,6 +15,7 @@
 
 import pytest
 import logging
+import botocore
 
 from e2e import service_marker
 from e2e.common.fixtures import (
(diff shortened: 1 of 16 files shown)


test/e2e/tests/test_data_quality_job_definition.py
```diff
@@ -34,9 +35,9 @@ def describe_sagemaker_data_quality_job_definition(
         return sagemaker_client.describe_data_quality_job_definition(
             JobDefinitionName=job_definition_name
         )
-    except BaseException:
+    except botocore.exceptions.ClientError as error:
         logging.error(
-            f"Could not find Data Quality Job Definition with name {job_definition_name}"
+            f"Could not find Data Quality Job Definition with name {job_definition_name}. Error {error}"
         )
```
akartsky, line 41: What if, for some reason, boto3 fails with a different exception other than `botocore.exceptions.ClientError`
then it will silently fail 

i think we should have one more `except` block below this 

Eg:
```
try:
  do some stuff 
except botocore.exceptions.ClientError as error:
  logging.error()
except:
  logging.error("something else went wrong")
```

surajkota: Specific type of exceptions will go to block, otherwise it will throw exception as usual

Wont fail silently

akartsky (approved):

PR #32, opened 2021-05-18
Add samples for each CRD

ryansteakley:
Added samples for each CRD for the sagemaker.services.k8s.aws controller.

PR #36, opened 2021-05-27
Add default values to the comparison checker

mbaijal:
Add default values to the comparison checker such that it does not pick up a diff and continue reconciling. 

### Testing:
1. Tested that local tests are passing using `make-kind-test` (except AdoptedResource test fails)
2. Checked the logs to ensure there are no extra diff (except HPO)

One final test run on the latest commit it in-progress.


pkg/resource/hyper_parameter_tuning_job/custom_delta.go
```diff
@@ -0,0 +1,22 @@
...
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+// Code generated by ack-generate. DO NOT EDIT.
```
surajkota, line 14: nit: not a generated file

pkg/resource/processing_job/custom_delta.go
```diff
@@ -0,0 +1,31 @@
...
+
+func customSetDefaults(
+	a *resource,
+	b *resource,
+) {
+	if a.ko.Spec.StoppingCondition == nil && b.ko.Spec.StoppingCondition != nil {
+		a.ko.Spec.StoppingCondition = &svcapitypes.ProcessingStoppingCondition{}
+	}
+
+	if a.ko.Spec.StoppingCondition.MaxRuntimeInSeconds == nil && b.ko.Spec.StoppingCondition.MaxRuntimeInSeconds != nil{
```
surajkota, line 28: for all these resources I think, `EnableManagedSpotTraining`,  `enableNetworkIsolation` and `enableInterContainerTrafficEncryption` will have default value false

pkg/resource/training_job/custom_delta.go
```diff
@@ -0,0 +1,43 @@
...
+
+	if a.ko.Spec.EnableNetworkIsolation == nil && b.ko.Spec.EnableNetworkIsolation != nil {
+		a.ko.Spec.EnableNetworkIsolation = b.ko.Spec.EnableNetworkIsolation
+	}
+
+	if a.ko.Spec.AlgorithmSpecification.EnableSageMakerMetricsTimeSeries == nil && b.ko.Spec.AlgorithmSpecification.EnableSageMakerMetricsTimeSeries != nil {
+		a.ko.Spec.AlgorithmSpecification.EnableSageMakerMetricsTimeSeries = b.ko.Spec.AlgorithmSpecification.EnableSageMakerMetricsTimeSeries
+	}
+
+	if a.ko.Spec.AlgorithmSpecification.EnableSageMakerMetricsTimeSeries == nil && b.ko.Spec.AlgorithmSpecification.EnableSageMakerMetricsTimeSeries != nil {
```
surajkota, line 36: `#32` and `#36` code duplication

pkg/resource/training_job/custom_delta.go
```diff
@@ -0,0 +1,43 @@
...
+		a.ko.Spec.AlgorithmSpecification.EnableSageMakerMetricsTimeSeries = b.ko.Spec.AlgorithmSpecification.EnableSageMakerMetricsTimeSeries
+	}
+
+	if a.ko.Spec.OutputDataConfig.KMSKeyID == nil && b.ko.Spec.OutputDataConfig.KMSKeyID != nil {
```
surajkota, line 40: this is so strange, why would sagemaker encrypt without user passing this

pkg/resource/transform_job/custom_delta.go
```diff
@@ -0,0 +1,56 @@
...
+		a.ko.Spec.DataProcessing.InputFilter = b.ko.Spec.DataProcessing.InputFilter
+	}
+	if a.ko.Spec.DataProcessing.JoinSource == nil && b.ko.Spec.DataProcessing.JoinSource != nil {
+		a.ko.Spec.DataProcessing.JoinSource = b.ko.Spec.DataProcessing.JoinSource
+	}
+	if a.ko.Spec.DataProcessing.OutputFilter == nil && b.ko.Spec.DataProcessing.OutputFilter != nil {
+		a.ko.Spec.DataProcessing.OutputFilter = b.ko.Spec.DataProcessing.OutputFilter
+	}
+
+	if a.ko.Spec.TransformOutput == nil && b.ko.Spec.TransformOutput != nil{
```
surajkota, line 45: can we replace `#45-53` by

```
if a.ko.Spec.TransformOutput == nil && b.ko.Spec.TransformOutput != nil {
		a.ko.Spec.TransformOutput = b.ko.Spec.TransformOutput
	}
```

pkg/resource/training_job/custom_delta.go
surajkota: I see its nil vs "" can you add a comment here?

pkg/resource/processing_job/custom_delta.go
mbaijal: I thought so too, but they don't show up in the logs, why do you think ?

pkg/resource/training_job/custom_delta.go
mbaijal: Fixed.

pkg/resource/training_job/custom_delta.go
mbaijal: done

pkg/resource/model/custom_delta.go
```diff
@@ -0,0 +1,36 @@
...
+
+func customSetDefaults(
+	a *resource,
+	b *resource,
+) {
+	if ackcompare.IsNil(a.ko.Spec.PrimaryContainer) && ackcompare.IsNotNil(b.ko.Spec.PrimaryContainer) {
+		a.ko.Spec.PrimaryContainer = &svcapitypes.ContainerDefinition{}
+	}
+
+	if ackcompare.IsNil(a.ko.Spec.PrimaryContainer.Mode) && ackcompare.IsNotNil(b.ko.Spec.PrimaryContainer.Mode) {
```
surajkota, line 29: we don't need to initialize primary container if it is nil because the server will only return value if its not nil. this should be

```
if ackcompare.IsNotNil(a.ko.Spec.PrimaryContainer) && ackcompare.IsNotNil(b.ko.Spec.PrimaryContainer) {
  if ackcompare.IsNil(a.ko.Spec.PrimaryContainer.Mode) && ackcompare.IsNotNil(b.ko.Spec.PrimaryContainer.Mode) 
  {
    a.ko.Spec.PrimaryContainer.Mode = b.ko.Spec.PrimaryContainer.Mode
  }
}
```

pkg/resource/processing_job/custom_delta.go
surajkota: lol was expecting behaviour to be the same across resources in sagemaker but looks like its not 😆

pkg/resource/model/custom_delta.go
mbaijal: I have added both cases - initialize if required, and then check again.

surajkota (approved):

PR #51, opened 2021-07-06
Create Failed terminal condition Feature Group Creating requeue 

BriannaRoskind:
Summary:
Set terminal condition for feature group in "Create Failed" state. Also factored out common custom_update_conditions code between feature group, monitoring schedule, and endpoint. Secondly, added requeue on Creating functionality for feature group in order to allow for the feature group status to be updated to Created.

Files Added:
- pkg/common/custom_set_terminal_state.go - Contains functionality of custom_update_conditions.go from previous endpoint / monitoring files (that is also reused with feature group).
- pkg/common/custom_set_synced_condition.go - Contains general functionality of customSetOutput from previous model package group hooks.go (that is also reused with feature group).
- pkg/common/custom_utilities.go - Contains helper function IsModifyingStatus that is used for both this PR and [PR #52](https://github.com/aws-controllers-k8s/sagemaker-controller/pull/52). Derived from previous model package group hooks.go.
- pkg/resource/feature_group/custom_update_conditions.go - added call to custom_set_terminal_state.go to set the terminal condition if the feature group "Create Failed" state is reached.
- pkg/resource/feature_group/hooks.go - Contains helper function isCreating, and ackrequeue variable requeueWaitWhileCreating which are referenced from (also new) resource manager customSetOutput function that is called post read One.

Files Edited:
- generator.yaml - Added custom update for featuregroup and updated captalization for custom update functions for endpoint and monitoring schedule to be consistent. Also included sdk_read_one_post_set_output hook calling to customSetOutput function.
- pkg/resource/endpoint/custom_update_conditions.go - factored out common code into custom_set_terminal_state.go
- pkg/resource/monitoring_schedule/custom_update_conditions.go - factored out common code into custom_set_terminal_state.go. Also replaced hardcoded string with library defined reference.
- pkg/resource/model_package_group/hooks.go - factored out common code into custom_set_synced_condition.go.

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,5 +1,5 @@
 ack_generate_info:
-  build_date: "2021-07-22T20:36:39Z"
+  build_date: "2021-07-23T02:02:51Z"
   build_hash: 6911f924191f5315747c8e81fbcbf29b4edc8cb1
   go_version: go1.16.4 linux/amd64
   version: v0.3.1
@@ -7,8 +7,8 @@ api_directory_checksum: 4906740ff8ff6ebfffbaea971f7e8aa5a858edd8
 api_version: v1alpha1
 aws_sdk_go_version: v1.38.11
 generator_config_info:
-  file_checksum: c05fa636159c1f79ab353fd790587a5b36f8b95b
+  file_checksum: 8e1b7e085d2ff8673e0d239e05e0a1403825c355
   original_file_name: generator.yaml
 last_modification:
   reason: API generation
-  timestamp: 2021-07-22 20:36:41.771709478 +0000 UTC
+  timestamp: 2021-07-23 02:02:53.613950676 +0000 UTC
--- a/generator.yaml
+++ b/generator.yaml
@@ -69,7 +69,7 @@ resources:
   Endpoint:
     reconcile: 
       requeue_on_success_seconds: 30
-    update_conditions_custom_method_name: customUpdateConditions
+    update_conditions_custom_method_name: CustomUpdateConditions
     exceptions:
       errors:
         404:
@@ -429,7 +429,7 @@ resources:
   MonitoringSchedule:
     reconcile: 
       requeue_on_success_seconds: 30
-    update_conditions_custom_method_name: customUpdateConditions
+    update_conditions_custom_method_name: CustomUpdateConditions
     exceptions:
       errors:
           404:
@@ -490,6 +490,7 @@ resources:
           operation: DescribeMonitoringSchedule
           path: LastMonitoringExecutionSummary
   FeatureGroup:
+    update_conditions_custom_method_name: CustomUpdateConditions
     exceptions:
       errors:
           404:
@@ -514,6 +515,8 @@ resources:
         template_path: common/sdk_delete_pre_build_request.go.tpl
       sdk_delete_post_request:
         template_path: common/sdk_delete_post_request.go.tpl
+      sdk_read_one_post_set_output:
+        code: rm.customSetOutput(&resource{ko})
     fields:
       FailureReason:
         is_read_only: true
--- a/pkg/common/custom_conditions.go
+++ b/pkg/common/custom_conditions.go
@@ -1,0 +1,74 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+// Use this file if conditions need to be updated based on the latest status
+// of endpoint which is not evident from API response
+
+package common
+
+import (
+	ackcondition "github.com/aws-controllers-k8s/runtime/pkg/condition"
+	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// SetSyncedCondition sets the ACK Synced Condition
+// status to true if the resource status exists but
+// is not in one of the given modifying statuses,
+// or false if the resource status is one of the
+// given modifying statuses.
+func SetSyncedCondition(
+	r acktypes.AWSResource,
+	latestStatus *string,
+	resourceName *string,
+	modifyingStatuses *[]string,
+) {
+	if latestStatus == nil {
+		return
+	}
+
+	msg := *resourceName + " is in " + *latestStatus + " status."
+	conditionStatus := corev1.ConditionTrue
+	if IsModifyingStatus(latestStatus, modifyingStatuses) {
+		conditionStatus = corev1.ConditionFalse
+	}
+
+	ackcondition.SetSynced(r, conditionStatus, &msg, nil)
+}
+
+// SetTerminalState sets conditions (terminal) on
+// a resource's supplied status conditions if the
+// latest status matches the terminal status.
+// It returns true if conditions are updated.
+func SetTerminalState(
+	r acktypes.AWSResource,
+	latestStatus *string,
+	resourceName *string,
+	terminalStatus string,
+) bool {
+	if latestStatus == nil || *latestStatus != terminalStatus {
+		return false
+	}
+
+	terminalCondition := ackcondition.Terminal(r)
+	if terminalCondition != nil && terminalCondition.Status == corev1.ConditionTrue {
+		// some other exception already put the resource in terminal condition
+		return false
+	}
+
+	// setting terminal condition since controller can no longer recover by retrying
+	terminalMessage := *resourceName + " status reached terminal state: " + terminalStatus + ". Check the FailureReason."
+	ackcondition.SetTerminal(r, corev1.ConditionTrue, &terminalMessage, nil)
+
+	return true
+}
--- a/pkg/resource/endpoint/custom_update_conditions.go
+++ b/pkg/resource/endpoint/custom_update_conditions.go
@@ -17,48 +17,24 @@
 package endpoint
 
 import (
-	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
 	svcapitypes "github.com/aws-controllers-k8s/sagemaker-controller/apis/v1alpha1"
-	"github.com/aws/aws-sdk-go/aws"
+	svccommon "github.com/aws-controllers-k8s/sagemaker-controller/pkg/common"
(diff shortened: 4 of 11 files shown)


pkg/common/custom_set_terminal_state.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// CustomUpdateConditions sets conditions (terminal) on supplied endpoint
+// it examines supplied resource to determine conditions.
+// It returns true if conditions are updated
+func SetTerminalState(
+	koStatusConditions *[]*ackv1alpha1.Condition,
+	latestStatus *string,
+	terminalStatus string,
+	terminalMessageAddress *string,
```
akartsky, line 31: remove `Address` in the variable name  
just keep it as `terminalErrorMessage`

pkg/resource/endpoint/custom_update_conditions.go
```diff
@@ -17,48 +17,26 @@
...
 import (
-	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
 	svcapitypes "github.com/aws-controllers-k8s/sagemaker-controller/apis/v1alpha1"
-	"github.com/aws/aws-sdk-go/aws"
+	customShared "github.com/aws-controllers-k8s/sagemaker-controller/pkg/common"
 	svcsdk "github.com/aws/aws-sdk-go/service/sagemaker"
-	corev1 "k8s.io/api/core/v1"
 )
 
+var terminalMessage = "endpoint status: Failed. Check FailureReason"
```
akartsky, line 25: Keep all the error messages in the same format

pkg/resource/endpoint/custom_update_conditions.go
```diff
@@ -17,48 +17,26 @@
...
-		}
-		ko.Status.Conditions = append(ko.Status.Conditions, terminalCondition)
-	}
-	terminalCondition.Status = corev1.ConditionTrue
-	terminalCondition.Message = aws.String("endpoint status: Failed. check FailureReason")
-
-	return true
+	terminalStatus := svcsdk.EndpointStatusFailed
+	koStatusConditions := &ko.Status.Conditions
+	terminalMessageAddress := &terminalMessage
```
akartsky, line 38: we don't need a new variable here  
you can just pass `&terminalMessage` to the function

pkg/resource/endpoint/custom_update_conditions.go
akartsky: Also usually variable names for constants are capitalized.  
Check the ack convention and follow the same

apis/v1alpha1/ack-generate-metadata.yaml
```diff
@@ -1,14 +1,14 @@
 ack_generate_info:
-  build_date: "2021-06-22T20:41:21Z"
-  build_hash: 38e309d07369b1e33250473084f91db3894c4379
-  go_version: go1.16.4 darwin/amd64
-  version: v0.2.3
-api_directory_checksum: 650887de24f94277e21771990d1391198cce93c3
+  build_date: "2021-07-06T02:04:40Z"
+  build_hash: 6911f924191f5315747c8e81fbcbf29b4edc8cb1
+  go_version: go1.16.4 linux/amd64
+  version: v0.3.1
```
mbaijal, line 5: A lot of these changes are overlapping between yours, Ryans and my PRs now.

generator.yaml
```diff
@@ -485,6 +485,48 @@ resources:
         from:
           operation: DescribeMonitoringSchedule
           path: LastMonitoringExecutionSummary
+  FeatureGroup:
+    # CustomUpdateConditions is used for setting terminal
```
mbaijal, line 489: Do we have comments in the generator.yaml at other places ? I would recommend adding the description to the method instead.

pkg/common/custom_set_terminal_state.go
```diff
@@ -0,0 +1,59 @@
...
+// of endpoint which is not evident from API response
+
+package common
+
+import (
+	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// CustomUpdateConditions sets conditions (terminal) on supplied endpoint
```
mbaijal, line 24: not just endpoint.

pkg/resource/endpoint/custom_update_conditions.go
mbaijal: Can we avoid the global string entirely ?

pkg/resource/endpoint/custom_update_conditions.go
mbaijal: or atleast use `svcsdk.EndpointStatusFailed` instead of Failed. Also, did we consider printing the failure reason here ?

pkg/resource/feature_group/custom_update_conditions.go
```diff
@@ -0,0 +1,39 @@
...
+
+package feature_group
+
+import (
+	svcapitypes "github.com/aws-controllers-k8s/sagemaker-controller/apis/v1alpha1"
+	customShared "github.com/aws-controllers-k8s/sagemaker-controller/pkg/common"
+	svcsdk "github.com/aws/aws-sdk-go/service/sagemaker"
+)
+
+var terminalMessage = "Feature group status: create failed"
```
mbaijal, line 22: Same comments as the custom_update for endpoint across all the resources here.

mbaijal (changes requested):

pkg/resource/endpoint/custom_update_conditions.go
surajkota: failure reason will be set in a separate field under status

surajkota (changes requested):
good that we are trying to clean up. May I suggest a slightly different approach to decouple the condition handling and checks for setting the condition itself

split this into two parts:
1. common pkg only handles get and set conditions - this can be extended to set other conditions like ResourceSynced as well. e.g. set/getSyncedCondition, set/getTerminalCondition
2. individual resource pkgs handle the checks related to the condition. e.g.
```
if resourceHasTerminalStatus(resource) {
  msg := "ABC is in '"+*resource.ko.Status.ABCStatus+"' status. Check FailureReason"
  common.setTerminalCondition(resource, corev1.ConditionTrue, &msg, nil)
  return true
}
```

```
func resourceHasTerminalStatus(r *resource) bool {
	if r.ko.Status.Status == nil || *latestStatus != "Failed" {
		return false
	}

	return true
}
```

This way the code can handle when there are multiple checks like,
- if resource is in CreateFailed/UpdateFailed status setTerimalCondition()
- if resource is in Creating/InProgress/Deleting etc state setResourceSyncedCondition()

surajkota:
Also, can you increase the scope of PR to handle `Creating` state as well?

surajkota:
FYI: 1. is already present to runtime repo. so you can just use it

https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/condition/condition.go

templates/feature_group/sdk_create_pre_build_request.go.tpl
```diff
@@ -0,0 +1,5 @@
+
+// If creating, requeue with wait until status becomes created.
+if isCreating(desired) {
```
surajkota, line 3: unreachable condition?

templates/feature_group/sdk_create_post_request.go.tpl
```diff
@@ -0,0 +1,10 @@
+
+// If creating, requeue with wait until status becomes created.
+if err == nil {
+        if foundResource, err := rm.sdkFind(ctx, desired); err != ackerr.NotFound {
```
surajkota, line 4: look at how other resources handle this. just need to set appropriate conditions post create

surajkota (changes requested):

akartsky (approved):

surajkota (changes requested):

pkg/common/custom_set_synced_condition.go
```diff
@@ -0,0 +1,42 @@
...
+	r resourceTypes.AWSResource,
+	latestStatus *string,
+	resourceName *string,
+	modifyingStatuses *[]string,
+) {
+	if latestStatus == nil {
+		return
+	}
+	msg := *resourceName + " is in " + *latestStatus + " status."
+	syncedCondition := corev1.ConditionTrue
```
surajkota, line 37: rename conditionStatus

pkg/common/custom_set_terminal_state.go
```diff
@@ -0,0 +1,60 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
surajkota, line 1: rename file to custom_conditions.go

pkg/resource/feature_group/hooks.go
```diff
@@ -0,0 +1,32 @@
...
+// permissions and limitations under the License.
+
+package feature_group
+
+import (
+	customShared "github.com/aws-controllers-k8s/sagemaker-controller/pkg/common"
+	svcsdk "github.com/aws/aws-sdk-go/service/sagemaker"
+)
+
+var resourceName = "Feature group"
```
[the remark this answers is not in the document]
BriannaRoskind: It is global because across the deleting and creating feature group pr's (#51 and #52), this variable is reused in two different functions in this hooks.go file: modifyingStatuses() and requeueUntilCanModify().

surajkota (changes requested):
I think there was a to-do in test to add resource side check for `Created`. Please address that. also, add assertions on sync condition

pkg/common/custom_conditions.go
```diff
@@ -0,0 +1,74 @@
...
+// permissions and limitations under the License.
+
+// Use this file if conditions need to be updated based on the latest status
+// of endpoint which is not evident from API response
+
+package common
+
+import (
+	ackcondition "github.com/aws-controllers-k8s/runtime/pkg/condition"
+	resourceTypes "github.com/aws-controllers-k8s/runtime/pkg/types"
```
surajkota, line 21: resourceTypes -> acktypes

pkg/common/custom_conditions.go
```diff
@@ -0,0 +1,74 @@
...
+
+	ackcondition.SetSynced(conditionManager, conditionStatus, &msg, nil)
+}
+
+// SetTerminalState sets conditions (terminal) on
+// a resource's supplied status conditions if the
+// latest status matches the terminal status.
+// It returns true if conditions are updated.
+func SetTerminalState(
+	conditionManager resourceTypes.AWSResource,
```
surajkota, line 54: same

pkg/common/custom_conditions.go
```diff
@@ -0,0 +1,74 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+// SetSyncedCondition sets the ACK Synced Condition
+// status to true if the resource status exists but
+// is not in one of the given modifying statuses,
+// or false if the resource status is one of the
+// given modifying statuses.
+func SetSyncedCondition(
+	conditionManager resourceTypes.AWSResource,
```
surajkota, line 31: rename to r?

surajkota (changes requested):
naming nits

Thanks. Good shape now

surajkota (approved):
/lgtm

PR #50, opened 2021-06-29
Feature group testing

BriannaRoskind:
Summary:
Generated Feature Group controller code and wrote passing integration tests for the Feature Group controller.

Files Added:
- test/e2e/tests/test_feature_group.py - Contains Create and Delete integration test for feature group
- test/e2e/resources/feature_group.yaml - yaml file for feature group creation in integration testing.
- pkg/resource/feature_group/custom_delta.go - Sets default spec values for feature group fields (DisableGlueTableCreation and S3StorageConfig) if the user did not define them.

Files Edited:
- test/e2e/service_bootstrap.py - Attached AmazonSageMakerFeatureStoreAccess Role Policy to the SageMaker Execution Role for testing purposes. (Feature Store requires this permission for the Feature Group creation and access)
- generator.yaml - Commented out feature group from the ignore list. Also added feature group exceptions, hooks, and fields.
- pkg/resource/feature_group/delta.go - added the CustumSetDefault() function from custom_delta.go to the newResourceDelta() function.

PR #52, opened 2021-07-07
Feature group delete failed requeue

BriannaRoskind:
This PR assumes feature group testing code from PR #50 is already in place, and that the feature group create failed terminal condition code from PR #51 is already in place.

Summary:
Added requeue on delete if the feature group "Delete Failed" state is reached. This is because the design decision was made to retry to delete if a delete ever fails / does not complete successfully.

Files Added:
- pkg/resource/feature_group/hooks.go - Contains helper function isDeleteFailed, and ackrequeue variable requeueWaitWhileDeleteFailed which are referenced from the feature group sdk_delete template files.
- pkg/common/custom_pre_build_delete_requeue_logic.go - contains functionality of requeueUntilCanModify from old model package group hooks.go.
- pkg/common/custom_utilities.go - Contains helper function IsModifyingStatus that is used for both this PR and [PR #51](https://github.com/aws-controllers-k8s/sagemaker-controller/pull/51). Derived from previous model package group hooks.go.

Files Edited:
- generator.yaml - Added feature group hooks for sdk_delete_pre_build_request sdk_delete_post_request.
- test/e2e/tests/test_feature_group.py - Removed delete failed TODO comments.
- pkg/resource/model_package_group/hooks.go - refactored to use common code from pkg/common directory.
- templates/common/sdk_delete_post_request.go.tpl - fixed indentation and missing line at EOF.
- templates/common/sdk_delete_pre_build_request.go.tpl - fixed indentation and missing line at EOF.

### DIFF
--- a/pkg/resource/feature_group/hooks.go
+++ b/pkg/resource/feature_group/hooks.go
@@ -1,0 +1,42 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package feature_group
+
+import (
+	"context"
+	"errors"
+	ackrequeue "github.com/aws-controllers-k8s/runtime/pkg/requeue"
+	customShared "github.com/aws-controllers-k8s/sagemaker-controller/pkg/common"
+	svcsdk "github.com/aws/aws-sdk-go/service/sagemaker"
+)
+
+var (
+	requeueWaitWhileDeleting = ackrequeue.NeededAfter(
+		errors.New("Still in the process of deleting."),
+		ackrequeue.DefaultRequeueAfterDuration,
+	)
+)
+
+// requeueUntilCanModify creates and returns an
+// ackrequeue error if a resource's latest status matches
+// any of the defined modifying statuses below.
+func (rm *resourceManager) requeueUntilCanModify(
+	ctx context.Context,
+	r *resource,
+) error {
+	latestStatus := r.ko.Status.FeatureGroupStatus
+	resourceName := "Feature group"
+	modifyingStatuses := []string{svcsdk.FeatureGroupStatusCreating, svcsdk.FeatureGroupStatusDeleting}
+	return customShared.ACKRequeueIfModifying(latestStatus, &resourceName, &modifyingStatuses)
+}
--- a/pkg/resource/feature_group/hooks.go
+++ b/pkg/resource/feature_group/hooks.go
@@ -25,7 +25,7 @@ var (
 	modifyingStatuses = []string{svcsdk.FeatureGroupStatusCreating,
 		svcsdk.FeatureGroupStatusDeleting}
 
-	resourceName = "FeatureGroup"
+	resourceName = resourceGK.Kind
 
 	requeueWaitWhileDeleting = ackrequeue.NeededAfter(
 		errors.New(resourceName+" is deleting."),
--- a/pkg/resource/feature_group/hooks.go
+++ b/pkg/resource/feature_group/hooks.go
@@ -17,16 +17,21 @@ import (
 	"context"
 	"errors"
 	ackrequeue "github.com/aws-controllers-k8s/runtime/pkg/requeue"
-	customShared "github.com/aws-controllers-k8s/sagemaker-controller/pkg/common"
+	svccommon "github.com/aws-controllers-k8s/sagemaker-controller/pkg/common"
 	svcsdk "github.com/aws/aws-sdk-go/service/sagemaker"
 )
 
 var (
 	requeueWaitWhileDeleting = ackrequeue.NeededAfter(
-		errors.New("Still in the process of deleting."),
+		errors.New("FeatureGroup is deleting."),
 		ackrequeue.DefaultRequeueAfterDuration,
 	)
 )
+
+var resourceName = "FeatureGroup"
+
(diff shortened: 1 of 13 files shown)


test/e2e/service_bootstrap.py
```diff
@@ -55,6 +55,11 @@ def create_execution_role() -> str:
     iam.attach_role_policy(
         RoleName=role_name, PolicyArn="arn:aws:iam::aws:policy/AmazonS3FullAccess"
     )
+    # This policy accommodates use of the feature group controller.
+    iam.attach_role_policy(
+        RoleName=role_name,
+        PolicyArn="arn:aws:iam::aws:policy/AmazonSageMakerFeatureStoreAccess"
```
akartsky, line 61: why do we need this ?

templates/feature_group/sdk_delete_pre_build_request.go.tpl
```diff
@@ -0,0 +1,4 @@
+// If a delete failed, requeue on delete.
```
akartsky, line 1: why do we need pre_build_request and post_request ?
isn't `sdk_delete_post_request.go.tpl ` enough ?

test/e2e/service_bootstrap.py
BriannaRoskind: I made the change of removing it from the feature group pr, but did not make the change here yet. I will remove this.

templates/feature_group/sdk_delete_pre_build_request.go.tpl
BriannaRoskind: Good suggestion! I looked more into it, as I had modeled this after a similar elastic-cache PR for requeue on "deleting" state, and concluded that for a delete failed state (rather than deleting) this pre_build_request should not be used. I went ahead and removed the pre_build_request hook and code. (We can eventually check this further in unit testing)

akartsky (approved):

templates/feature_group/sdk_delete_post_request.go.tpl
```diff
@@ -0,0 +1,9 @@
+// If a delete failed, requeue on delete.
+if err == nil {
+        if foundResource, err := rm.sdkFind(ctx, r); err != ackerr.NotFound {
+                if isDeleteFailed(foundResource) {
+                        return requeueWaitWhileDeleteFailed
```
surajkota, line 5: why is the requeue on DeleteFailed status?

write out what we are trying to achieve in the PR description. It should help you understand

surajkota (changes requested):
left comments

will review after other PRs are merged

surajkota (changes requested):
FeatureGroup has the following states:
`'Creating'|'Created'|'CreateFailed'|'Deleting'|'DeleteFailed'`

1. Resource can be in any state when delete is called. If the resource is currently being modified on the service side(Creating, Deleting) sagemakerSDK.delete should not be called. Instead, the controller needs to requeue until resource can be modified/deleted

2. Once delete is called, we need to call sdkRead again so that resource can get patched with `Deleting` or `DeleteFailed` status and controller to requeue until the resource is completely deleted from the service side

The current revision of PR partially handles 2 but not 1. Partially 2 because it does not account for Deleting status. 

Note: 1 is based on the assumption that feature group behaviour is same as other resources and SageMaker will throw an error if you try to delete a resource when its being modified by server

-----

Question: Do we even need a check `isDeleteFailed()`? if the resource is found, isn't it a sufficient condition for requeue?
lmk if I am missing something

-----

FYI: I see this as final PR which should handle everything related to Deletion and not only DeleteFailed

Also, it will be good to update runtime and codegen 0.5.1 in this PR since patching after delete is implemented in that version but it's not a blocker

nit: clear the TODO in the test

pkg/resource/feature_group/hooks.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+	"errors"
+	ackrequeue "github.com/aws-controllers-k8s/runtime/pkg/requeue"
+	customShared "github.com/aws-controllers-k8s/sagemaker-controller/pkg/common"
+	svcsdk "github.com/aws/aws-sdk-go/service/sagemaker"
+)
+
+var (
+	requeueWaitWhileDeleting = ackrequeue.NeededAfter(
+		errors.New("Still in the process of deleting."),
```
surajkota, line 26: nit: Featuregroup is deleting

pkg/resource/feature_group/hooks.go
```diff
@@ -0,0 +1,43 @@
...
+
+// requeueUntilCanModify creates and returns an
+// ackrequeue error if a resource's latest status matches
+// any of the defined modifying statuses below.
+func (rm *resourceManager) requeueUntilCanModify(
+	ctx context.Context,
+	r *resource,
+) error {
+	latestStatus := r.ko.Status.FeatureGroupStatus
+	resourceName := "Feature group"
```
surajkota, line 39: FeatureGroup

surajkota (changes requested):
few comments

pkg/resource/model_package_group/hooks.go
```diff
@@ -29,46 +30,32 @@ var (
 		ackrequeue.DefaultRequeueAfterDuration,
 	)
 )
+var resourceName = "Model package group"
```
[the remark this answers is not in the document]
BriannaRoskind: It is global because across the deleting and creating feature group pr's (#51 and #52), this variable is reused in two different functions in this hooks.go file: modifyingStatuses() and requeueUntilCanModify().

pkg/resource/feature_group/hooks.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// requeueUntilCanModify creates and returns an
+// ackrequeue error if a resource's latest status matches
+// any of the defined modifying statuses below.
+func (rm *resourceManager) requeueUntilCanModify(
+	ctx context.Context,
+	r *resource,
+) error {
+	latestStatus := r.ko.Status.FeatureGroupStatus
```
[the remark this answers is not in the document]
BriannaRoskind: I added the latestStatus nil check in the ACKRequeueIfModifying function.

surajkota (approved):
/approve
/lgtm

PR #56, opened 2021-07-15
Notebook Instance (Create/Delete) test

ananth102:
This is a pull request for the Notebook Instance, this PR contains code that will enable the creation/deletion of Notebook Instances with ACK.

Create operation: Creates a notebook instance and sets the resource synced status to false until the NotebookInstance reaches the InService state(custom_set_output.go). Also fills in any unfilled values to prevent an update after creation(custom_set_delta.go).

Delete Operation: Stops the notebook instance and deletes it. Notebook has to be in the “Stopped” state in order for it to be deleted.

Integration test checks if creation/deletion are working correctly.

Custom code:
pkg/resource/notebook_instance/custom_sdk_api.go : File containing functions that call the sagemaker API
pkg/resource/notebook_instance/custom_set_delta.go : Fills in default values to prevent an update after creation
pkg/resource/notebook_instance/hooks.go: Handles re queuing and setting resource synced conditions.
templates/notebook_instance/sdk_delete_pre_build_request.go.tpl : Stops the notebook and re queues on any state other than Stopped/Failed.

Integration test files:

test/e2e/resources/notebook_instance.yaml : Resource spec used for test.
test/e2e/tests/test_notebook_instance.py : Integration test for notebook, creates and deletes the notebook and verifies that states were synced.

PR #61, opened 2021-07-28
Notebook update code + updated integration test

ananth102:
Notebook Instance Update PR

This is a PR that adds the update operation to the notebook operator and also contains a sample.

How update works:

1. Controller stops the Notebook if its not Stopped already.(Sets a status field "StoppedByControllerMETA") if it did stop the notebook.
2. Controller requeues during intermediate "Stopping" state until it reaches "Stopped"
3. Controller updates the notebook and also sets the StoppedByControllerMETA to "UpdateTriggered" if its not nil.
4. After the update the controller starts the Notebook if StoppedByControllerMETA is "UpdateTriggered".
5. The controller resets the StoppedByControllerMETA status field.

Note:

The update operation is supposed to start the notebook after an update if it was not in a stopped state and keep the notebook in the stopped state if it was in the stopped state prior to an update.

We use a status field StoppedByControllerMETA to keep track of whether the notebook was stopped by a controller and an annotation done_updating to keep track of whether an update has finished. 

Why is StoppedByControllerMETA a status and not an annotation?
After the controller stops a Notebook Instance, it can’t update the notebook because it will usually be in the stopping state and will stay in “Stopping” for at least a minute. Because of this an error is returned as the notebook can only be updated when it is in the stopped state. If we dont return an error the spec gets updated and the controller wont be able to detect a delta.
Currently it is impossible for the controller to update an annotation if an update call has failed. If an update fails the controller patches the resource status and not the metadata(https://github.com/aws-controllers-k8s/runtime/blob/055b089c6a508317ad4bb57ce037dc55061e1829/pkg/runtime/reconciler.go#L235). We can’t also not return an error because that would patch the spec too(which would end up with the controller not calling update).

Note:
Even after 0.7.1, you cant update annotations(metadata) without also updating the spec, we cannot update the spec when we stop the notebook as that would prevent the update from occurring(runtime will not detect a diff).(https://github.com/aws-controllers-k8s/runtime/blob/d7b6e45c7ba9e54664b49f681747abf275a232f9/pkg/runtime/reconciler.go#L587)

Files added:

custom_sdk_api.go: Handles api calls to Sagemaker

Hooks.go: Contains code to requeue and also customsetoutputDescribe which is used to set the resource synced conditions and also used to start the Notebook after an update(if it was not stopped before the update).

notebook_instance/sdk_read_one_post_set_output.go.tpl:
Calls custom_set_describe
notebook_instance/sdk_read_one_pre_set_output.go.tpl. 
sdk_read_one_pre_set_output.go.tpl sets the LifecycleConfigName spec field(code generator does not generate it).

notebook_instance/sdk_update_pre_build_request.go.tpl:
Requeues if in the stopping/updating/pending state. If the Notebook is in the InService state stopNotebookInstance is called, if it does not return an error, the status field StoppedByControllerMETA is set to true.

notebook_instance/sdk_update_post_set_output.go.tpl:
Sets the is_updating status and sets the resource synced condition to false so the controller requeues.
Covered in Create/Delete:
https://github.com/aws-controllers-k8s/sagemaker-controller/pull/56

Removed a check for assert for UpdateTriggered because it makes the test much more flakey because update can take as little as a minute at times

————————————————————————————————————

Custom code:
pkg/resource/notebook_instance/custom_set_delta.go : Fills in default values to prevent an update after creation
pkg/resource/notebook_instance/hooks.go :  Contains code to requeue and set sync conditions, also contains custom_set_output_describe that starts a Notebook(after an update).
templates/notebook_instance/sdk_delete_pre_build_request.go.tpl : Stops the notebook and re queues on any state other than Stopped/Failed.


test/e2e/tests/test_notebook_instance.py
```diff
@@ -0,0 +1,135 @@
...
+        notebook_instance_name = resource["spec"].get("notebookInstanceName",None)
+        assert notebook_instance_name is not None
+
+        notebook_description = get_notebook_instance(notebook_instance_name)
+        assert notebook_description["NotebookInstanceStatus"] == "Pending"
+
+        #wait for the resource to go to the InService state and make sure the operator is synced with sagemaker.
+        self._assert_notebook_status_in_sync(notebook_instance_name,reference,"InService")
+
+        spec["spec"]["volumeSizeInGB"] = 7
```
ryansteakley, line 121: Is this where you are testing the update? Can you add a comment here to clarify

test/e2e/tests/test_notebook_instance.py
```diff
@@ -0,0 +1,135 @@
...
+        k8s.patch_custom_resource(reference,spec)
+
+        self._assert_notebook_status_in_sync(notebook_instance_name,reference,"InService")
+
+        resource = k8s.wait_resource_consumed_by_controller(reference)
+        assert resource is not None
+
+        latest_notebook = get_notebook_instance(notebook_instance_name)
```
ryansteakley, line 129: Can you also do a resource side check for this as well

test/e2e/tests/test_notebook_instance.py
```diff
@@ -0,0 +1,135 @@
...
+        assert(latest_notebook["VolumeSizeInGB"] == 7)
+
+
+        # Delete the k8s resource.
+        _, deleted = k8s.delete_custom_resource(reference, 11, 30)
```
ryansteakley, line 134: Can you do a service side check to verify its deleted as well

test/e2e/tests/test_notebook_instance.py
```diff
@@ -0,0 +1,135 @@
...
+        )
+    def testCreateAndDelete(self,notebook_instance):
+        (reference, resource, spec) = notebook_instance
+        assert k8s.get_resource_exists(reference)
+
+        notebook_instance_name = resource["spec"].get("notebookInstanceName",None)
+        assert notebook_instance_name is not None
+
+        notebook_description = get_notebook_instance(notebook_instance_name)
+        assert notebook_description["NotebookInstanceStatus"] == "Pending"
```
ryansteakley, line 116: can we also do self._assert_notebook_status_in_sync(notebook_instance_name,reference,"Pending") or will this cause flakyness as the time in pending may be too short?

test/e2e/tests/test_notebook_instance.py
```diff
@@ -0,0 +1,135 @@
...
+import random
+
+RESOURCE_PLURAL = "notebookinstances"
+RESOURCE_PREFIX = "nb"
+RESOURCE_SPEC_FILE = "notebook_instance"
+
+
+@pytest.fixture(scope="function")
+def notebook_instance():
+    resource_name = RESOURCE_PREFIX + str(random.randint(0, 10000))
```
ryansteakley, line 42: Is there a reason for this str(random.randint(0, 10000)) and not using something like [this](https://github.com/aws-controllers-k8s/sagemaker-controller/blob/bb7eb30d6802ba6caf7e644f216e0c645bef527e/test/e2e/tests/test_trainingjob.py#L39)?

pkg/resource/notebook_instance/sdk.go
```diff
@@ -0,0 +1,576 @@
...
+	}
+
+	rm.customPreDelete(r)
+	input, err := rm.newDeleteRequestPayload(r)
+	if err != nil {
+		return nil, err
+	}
+	var resp *svcsdk.DeleteNotebookInstanceOutput
+	_ = resp
+	resp, err = rm.sdkapi.DeleteNotebookInstanceWithContext(ctx, input)
```
ryansteakley, line 441: Do we also need something like [this ](https://github.com/aws-controllers-k8s/sagemaker-controller/blob/bb7eb30d6802ba6caf7e644f216e0c645bef527e/pkg/resource/model_package/sdk.go#L978) here?

test/e2e/tests/test_notebook_instance.py
ananth102: Didn't mean to have that file here. This test takes around 13 minutes so I feel its best to test update through a unit test.

test/e2e/tests/test_notebook_instance.py
ananth102: Mentioned it in the other PR but it's because the Notebook does not support that kind of name.

surajkota:
Runtime 0.9.2 was released with bug fixes and all the things you might need for annotations. Can you try it out?

surajkota:
Also, please open a separate PR for samples

ananth102:
Update:
It turns out that runtimes above 0.8.0 do not support updating metadata(annotations) during readOne. This operator does update annotations during readOne in order to reset StoppedByAck(determines whether the controller should start after an update true=start false=stop ) and isUpdating(stores if the controller entered updating). Without those two annotations being updated the controller does not behave correctly. As a result I've converted both variables into status fields.

Based on https://github.com/aws-controllers-k8s/runtime/commit/1b10a4268ff6b3d06ef2d776c6aa70f2e38185a3, this seems to be a design decision and not a bug. 
"Removes the `resourceReconciler.patchResource` method entirely (which
  patched both Metadata/Spec as well as Status) and ensures that the
  `createResource`, `updateResource` and `deleteResource` methods on
  `resourceReconciler` only ever call `patchResourceMetadataAndSpec`
  (since `patchResourceStatus` is now always called by the
  `HandleReconcileError` wrapper)"

mbaijal:
> * Controller updates the notebook and also sets a done_updating annotation to indicate that it has started updating, if there is an error it tries again when the notebook reaches the failed state.

Does this indicate "started updating" or "done updating" ?

pkg/resource/notebook_instance/hooks.go
```diff
@@ -0,0 +1,92 @@
...
+
+// customSetOutput sets the ack syncedCondition depending on
+// whether the latest status of the resource is one of the
+// defined modifyingStatuses.
+func (rm *resourceManager) customSetOutput(r *resource) {
+	latestStatus := r.ko.Status.NotebookInstanceStatus
+	svccommon.SetSyncedCondition(r, latestStatus, &resourceName, &modifyingStatuses)
+}
+
+func (rm *resourceManager) customSetOutputDescribe(r *resource,
```
mbaijal, line 64: @mbaijal Review this method again.

samples/notebook_instance/my-notebook-instance.yaml
```diff
@@ -0,0 +1,17 @@
+
```
mbaijal, line 1: TODO for team: Samples are named differently across resources.

samples/notebook_instance/my-notebook-instance.yaml
```diff
@@ -0,0 +1,17 @@
...
+  notebookInstanceName: my-notebook
+  roleARN: <YOUR SAGEMAKER ROLE ARN>
+  volumeSizeInGB: 6  
+  directInternetAccess: Enabled
+  rootAccess: Enabled
+  acceleratorTypes:
+    - ml.eia1.medium
+  lifecycleConfigName: sample-lifecycle
+  defaultCodeRepository: <Your git code repository>
+
```
mbaijal, line 17: Most YAMLs we have do not have an extra empty line, so could you remove it.

samples/notebook_instance/my-notebook-instance.yaml
```diff
@@ -0,0 +1,17 @@
...
+spec:
+  instanceType: ml.t2.medium
+  notebookInstanceName: my-notebook
+  roleARN: <YOUR SAGEMAKER ROLE ARN>
+  volumeSizeInGB: 6  
+  directInternetAccess: Enabled
+  rootAccess: Enabled
+  acceleratorTypes:
+    - ml.eia1.medium
+  lifecycleConfigName: sample-lifecycle
```
mbaijal, line 15: is this something you need to create separately ? If yes, could you mention in the README.

test/e2e/tests/test_notebook_instance.py
```diff
@@ -0,0 +1,169 @@
...
+        assert notebook_description["NotebookInstanceStatus"] == "Pending"
+
+        assert k8s.wait_on_condition(reference, "ACK.ResourceSynced", "False")
+        self._assert_notebook_status_in_sync(notebook_instance_name,reference,"Pending")
+
+        #wait for the resource to go to the InService state and make sure the operator is synced with sagemaker.
+        self._assert_notebook_status_in_sync(notebook_instance_name,reference,"InService")
+        assert k8s.wait_on_condition(reference, "ACK.ResourceSynced", "True")
+
+        #Update test
```
mbaijal, line 148: Could you test update on a field that is not a part of the create request such as - DisassociateLifecycleConfig

mbaijal: Also, could you move this to a separate test such as here - https://github.com/aws-controllers-k8s/sagemaker-controller/blob/main/test/e2e/tests/test_endpoint.py#L304

mbaijal (changes requested):

samples/notebook_instance/my-notebook-instance.yaml
ananth102: mentioned it in create

surajkota (changes requested):
Left comments
Havent finished reviewing `customSetOutputDescribe` method but its hard to understand as well. Please add comments

templates/notebook_instance/sdk_read_one_pre_set_output.go.tpl
```diff
@@ -0,0 +1,15 @@
+//TODO: Code generator wont autogenerate the line below, take this if statement out when it does.
```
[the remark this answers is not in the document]
ananth102: It's seems like a bug with code generator.

templates/notebook_instance/sdk_read_one_post_set_output.go.tpl
```diff
@@ -0,0 +1,7 @@
+modified_status := rm.customSetOutputDescribe(r, ko)
+//TODO: Take this out if runtime supports updating annotations when an error is returned.
+if !modified_status{
+  //covers a scenario where the code generator generates code to set StoppedByAck
+  ko.Status.StoppedByAck = &tmp
```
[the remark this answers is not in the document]
ananth102: nice catch!

surajkota (commented):
What is the story around Failed status?

- Can you call an update if NotebookStatus is Failed?
- How do we handle the case where InService -> Updating -> Failed

if startNotebook is called and aws is out of capacity, does it throw an error in the response, or is that async?

ananth102:
@surajkota 
> What is the story around Failed status?
> 
>     * Can you call an update if NotebookStatus is Failed?

Yes you can call an update if the Notebook when the Notebook is in the Failed state, that what this operator currently does.

> 
>     * How do we handle the case where InService -> Updating -> Failed
> 

We try to update again in that case.
> 
> if startNotebook is called and aws is out of capacity, does it throw an error in the response, or is that async?
If you mean ResourceLimitExceeded then it goes into a terminal condition. Need to look into the aws thing, but we should throw an error if start fails.

The only time when the Notebook is left in the Failed state is when it's lifecycle configuration fails. In that case there no point starting the Notebook back up(because it's going to fail again). The user can either delete or update the Notebook in that case.

generator.yaml
```diff
@@ -621,6 +621,74 @@ resources:
...
+          name: STATUS
+        from:
+          operation: DescribeNotebookInstance
+          path: NotebookInstanceStatus
+      Url:
+        is_read_only: true
+        from:
+          operation: DescribeNotebookInstance
+          path: Url
+        print:
```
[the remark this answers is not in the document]
ananth102: I'll take it out.

pkg/resource/notebook_instance/custom_sdk_api.go
```diff
@@ -0,0 +1,27 @@
...
+	input := &svcsdk.StopNotebookInstanceInput{}
+	input.SetNotebookInstanceName(*r.ko.Spec.NotebookInstanceName)
+
+	// Stop Notebook Instance does not return a response
+	_, err := rm.sdkapi.StopNotebookInstance(input)
+	rm.metrics.RecordAPICall("STOP", "StopNotebookInstance", err)
+	return err
+}
+
+func (rm *resourceManager) startNotebookInstance(r *resource) {
```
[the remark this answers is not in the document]
ananth102: done

pkg/resource/notebook_instance/hooks.go
```diff
@@ -0,0 +1,92 @@
...
+) error {
+	notebookStatus := r.ko.Status.NotebookInstanceStatus
+	return svccommon.RequeueIfModifying(notebookStatus, &resourceName, &modifyingStatuses)
+}
+
+func isNotebookUpdating(r *resource) bool {
+	if r.ko.Status.NotebookInstanceStatus == nil {
+		return false
+	}
+	notebookInstanceStatus := r.ko.Status.NotebookInstanceStatus
```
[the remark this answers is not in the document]
ananth102: it was, getting rid of it

pkg/resource/notebook_instance/hooks.go
```diff
@@ -0,0 +1,92 @@
...
+	if ko.Status.IsUpdating != nil && *ko.Status.IsUpdating == "true" {
+		if notebook_state != svcsdk.NotebookInstanceStatusStopped {
+			return false //we want to keep requeing until update finishes
+		}
+		//TODO: Use annotations instead of status once the runtime supports updating metadata
+		if ko.Status.StoppedByAck != nil && *ko.Status.StoppedByAck == "true" {
+			rm.startNotebookInstance(r)
+			ko.Status.IsUpdating = aws.String("false")
+			ko.Status.StoppedByAck = aws.String("false")
+			return true //dont want to set resource synced to true pre maturely.
```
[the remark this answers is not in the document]
ananth102: got rid of it

pkg/resource/notebook_instance/sdk.go
```diff
@@ -0,0 +1,620 @@
...
+		if err := rm.stopNotebookInstance(latest); err != nil {
+			return latest, err
+		} else {
+			stopped_by_ack = true
+		}
+	}
+
+	//TODO: Take this out if the runtime supports updating annotations if an error is returned and use annotations for this.
+	if stopped_by_ack {
+		latest.ko.Status.StoppedByAck = aws.String("true")
```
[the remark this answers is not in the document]
ananth102: took it out

templates/notebook_instance/sdk_read_one_pre_set_output.go.tpl
```diff
@@ -0,0 +1,15 @@
...
+if resp.NotebookInstanceLifecycleConfigName != nil {
+	ko.Spec.LifecycleConfigName = resp.NotebookInstanceLifecycleConfigName
+} else {
+	ko.Spec.LifecycleConfigName = nil
+}
+//TODO: Take this out if runtime supports updating annotations during ReadOne
+tmp := ""
```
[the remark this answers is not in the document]
ananth102: took it out after the meeting

templates/notebook_instance/sdk_update_post_set_output.go.tpl
```diff
@@ -0,0 +1,4 @@
+//TODO: Replace the is_updating status with an annotation if the runtime can update annotations after a readOne call.
+ko.Status.IsUpdating = aws.String("true")
```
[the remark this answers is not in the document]
ananth102: put it in a method

generator.yaml
```diff
@@ -621,6 +621,74 @@ resources:
...
+    hooks:
+      sdk_delete_pre_build_request:
+        template_path: notebook_instance/sdk_delete_pre_build_request.go.tpl
+      sdk_delete_post_request:
+        template_path: common/sdk_delete_post_request.go.tpl
+      delta_pre_compare:
+        code: customSetDefaults(a, b)
+      sdk_read_one_post_set_output: 
+        template_path: notebook_instance/sdk_read_one_post_set_output.go.tpl
+      sdk_read_one_pre_set_output:
```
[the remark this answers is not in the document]
ananth102: done

test/e2e/tests/test_notebook_instance.py
```diff
@@ -0,0 +1,193 @@
...
+        notebook_instance_name = resource["spec"].get("notebookInstanceName", None)
+
+        # Update test
+        spec["spec"]["volumeSizeInGB"] = 7
+        k8s.patch_custom_resource(reference, spec)
+        self._assert_notebook_status_in_sync(
+            notebook_instance_name, reference, "InService"
+        )
+
+        resource = k8s.wait_resource_consumed_by_controller(reference)
```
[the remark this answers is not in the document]
ananth102: took that line out

test/e2e/tests/test_notebook_instance.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def updateTest(self, notebook_instance):
+        (reference, resource, spec) = notebook_instance
+        notebook_instance_name = resource["spec"].get("notebookInstanceName", None)
+
+        # Update test
+        spec["spec"]["volumeSizeInGB"] = 7
+        k8s.patch_custom_resource(reference, spec)
+        self._assert_notebook_status_in_sync(
+            notebook_instance_name, reference, "InService"
```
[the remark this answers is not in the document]
ananth102: sometimes update can take as little as 10 seconds.

pkg/resource/notebook_instance/hooks.go
ananth102: added a comment

pkg/resource/notebook_instance/hooks.go
```diff
@@ -0,0 +1,92 @@
...
+// whether the latest status of the resource is one of the
+// defined modifyingStatuses.
+func (rm *resourceManager) customSetOutput(r *resource) {
+	latestStatus := r.ko.Status.NotebookInstanceStatus
+	svccommon.SetSyncedCondition(r, latestStatus, &resourceName, &modifyingStatuses)
+}
+
+func (rm *resourceManager) customSetOutputDescribe(r *resource,
+	ko *svcapitypes.NotebookInstance) bool {
+	notebook_state := *ko.Status.NotebookInstanceStatus // Get the Notebook State
```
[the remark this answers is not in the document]
ananth102: done

samples/notebook_instance/my-notebook-instance.yaml
ananth102: Moved this here https://github.com/aws-controllers-k8s/sagemaker-controller/pull/81

test/e2e/tests/test_notebook_instance.py
ananth102: Handled in #82

test/e2e/tests/test_notebook_instance.py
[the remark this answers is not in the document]
ananth102: Nevermind it's best to check for updating. Also added a check for UpdateTriggered.

templates/notebook_instance/sdk_read_one_post_set_output.go.tpl
```diff
@@ -0,0 +1,10 @@
+// NotebookInstanceLifecycleConfigName does not get generated by the code generator.
```
surajkota, line 1: why?
add this:
because the field name in spec vs Describe response do not match

pkg/resource/notebook_instance/hooks.go
```diff
@@ -48,3 +52,33 @@ func (rm *resourceManager) customSetOutput(r *resource) {
...
+	return nil
+}
+
+// customSetOutputUpdate sets the StoppedByControllerMeta field to UpdateTriggered if it was in UpdatePending before.
+func (rm *resourceManager) customSetOutputUpdate(ko *svcapitypes.NotebookInstance, latest *resource) {
+	//TODO: Replace the StoppedByControllerMeta status with an annotation if the runtime can update annotations after a readOne call.
+	if latest.ko.Status.StoppedByControllerMetadata != nil && *latest.ko.Status.StoppedByControllerMetadata == "UpdatePending" {
+		ko.Status.StoppedByControllerMetadata = aws.String("UpdateTriggered")
+	}
+	ackcond.SetSynced(&resource{ko}, corev1.ConditionFalse, nil, aws.String("Notebook is currenty updating"))
```
surajkota, line 83: nit: use `svccommon` instead

`svccommon.SetSyncedCondition(&resource{ko}, svcsdk.NotebookInstanceStatusUpdating, &resourceName, &modifyingStatuses)`

pkg/resource/notebook_instance/hooks.go
```diff
@@ -48,3 +52,33 @@ func (rm *resourceManager) customSetOutput(r *resource) {
 	latestStatus := r.ko.Status.NotebookInstanceStatus
 	svccommon.SetSyncedCondition(r, latestStatus, &resourceName, &modifyingStatuses)
 }
+
+// CustomSetOutputDescribe has three functions
+// 1. Setting the resource synced condition during sdk.Find
+// 2. Starting the Notebook if the StoppedByController field is set to UpdateTriggerd
+// 3. Reseting the StoppedByControllerMetadata field.
```
surajkota, line 59: nit: Reseting the StoppedByControllerMetadata field if notebook is started

pkg/resource/notebook_instance/hooks.go
```diff
@@ -48,3 +52,33 @@ func (rm *resourceManager) customSetOutput(r *resource) {
...
+func (rm *resourceManager) customSetOutputDescribe(r *resource,
+	ko *svcapitypes.NotebookInstance) error {
+	notebookStatus := *ko.Status.NotebookInstanceStatus // Get the Notebook State
+	if notebookStatus == svcsdk.NotebookInstanceStatusStopped {
+		if ko.Status.StoppedByControllerMetadata != nil && *ko.Status.StoppedByControllerMetadata == "UpdateTriggered" {
+			err := rm.startNotebookInstance(r)
```
surajkota, line 65: nit: if you use &resource{ko} here, customSetOutputDescribe will only need one argument

surajkota (changes requested):
Nice, this version is much cleaner

Left a few comments, mostly nits

helm/crds/sagemaker.services.k8s.aws_notebookinstances.yaml
```diff
@@ -0,0 +1,249 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1
```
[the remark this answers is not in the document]
ananth102: The integration tests will fail if I don't include them.

generator.yaml
```diff
@@ -621,6 +621,74 @@ resources:
...
+          operation: DescribeNotebookInstance
+          path: NotebookInstanceStatus
+      Url:
+        is_read_only: true
+        from:
+          operation: DescribeNotebookInstance
+          path: Url
+        print:
+          name: URL
+      stopped_by_ack:
```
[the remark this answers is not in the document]
surajkota: Add a comment saying, using this way since code generator only has the ability to create field types corresponding to a shape in operation. Ideally we want these fields to be boolean instead of string

test/e2e/tests/test_notebook_instance.py
```diff
@@ -123,10 +126,9 @@ def _assert_notebook_status_in_sync(
             == expected_status
         )
 
-    def testCreateAndDelete(self, notebook_instance):
-        (reference, resource) = notebook_instance
+    def create_notebook_test(self, notebook_instance):
+        (reference, resource, spec) = notebook_instance
         assert k8s.get_resource_exists(reference)
-        assert k8s.get_resource_arn(resource) is not None
```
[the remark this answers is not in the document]
ananth102: probably got removed in some merge, put it back

templates/notebook_instance/sdk_update_pre_build_request.go.tpl
```diff
@@ -0,0 +1,16 @@
+if err = rm.requeueUntilCanModify(ctx, latest); err != nil {
+	return latest, err
+}
+latestStatus := latest.ko.Status.NotebookInstanceStatus
+// The Notebook Instance is stopped and the StoppedByControllerMetadata status is
+// set to UpdatePending.
+if latestStatus != nil &&
+ *latestStatus == svcsdk.NotebookInstanceStatusInService {
+	if err := rm.stopNotebookInstance(latest); err != nil {
+		return latest, err
```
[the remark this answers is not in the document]
ananth102: done

templates/notebook_instance/sdk_delete_pre_build_request.go.tpl
```diff
@@ -1,3 +1,4 @@
+// This will avoid exponential backoff
```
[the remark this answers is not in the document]
ananth102: removed it

templates/notebook_instance/sdk_read_one_post_set_output.go.tpl
ananth102: added it

pkg/resource/notebook_instance/hooks.go
ananth102: done

pkg/resource/notebook_instance/hooks.go
```diff
@@ -48,3 +50,32 @@ func (rm *resourceManager) customSetOutput(r *resource) {
 	latestStatus := r.ko.Status.NotebookInstanceStatus
 	svccommon.SetSyncedCondition(r, latestStatus, &resourceName, &modifyingStatuses)
 }
+
+// CustomSetOutputDescribe has three functions
+// 1. Setting the resource synced condition during sdk.Find
+// 2. Starting the Notebook if the StoppedByController field is set to UpdateTriggerd
+// 3. Reseting the StoppedByControllerMetadata field if the Notebook starts.
+func (rm *resourceManager) customSetOutputDescribe(r *resource) error {
+	notebookStatus := *r.ko.Status.NotebookInstanceStatus // Get the Notebook State
```
surajkota, line 59: https://github.com/aws-controllers-k8s/sagemaker-controller/pull/61#pullrequestreview-731151817

check other places

test/e2e/tests/test_notebook_instance.py
```diff
@@ -146,7 +156,57 @@ def testCreateAndDelete(self, notebook_instance):
...
+        # TODO: Replace with annotations once runtime can update annotations in readOne.
+        resource = k8s.get_resource(reference)
+        assert resource["status"]["stoppedByControllerMetadata"] == "UpdatePending"
+
+        self._assert_notebook_status_in_sync(
+            notebook_instance_name,
+            reference,
+            "Updating",
+            wait_periods=40,
+            period_length=15,
```
surajkota, line 179: 30
set wait_period accordingly

surajkota (changes requested):

templates/notebook_instance/sdk_update_pre_build_request.go.tpl
```diff
@@ -0,0 +1,17 @@
...
+// The Notebook Instance is stopped and the StoppedByControllerMetadata status is
+// set to UpdatePending.
+if latestStatus != nil &&
+ *latestStatus == svcsdk.NotebookInstanceStatusInService {
+	if err := rm.stopNotebookInstance(latest); err != nil {
+		return nil, err
+	} else {
+		//TODO: Replace with annotations once rutime supports it.
+		latest.ko.Status.StoppedByControllerMetadata = aws.String("UpdatePending")
+	  return latest, requeueWaitWhileStopping
```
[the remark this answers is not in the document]
ananth102: We want to return latest to update the status field.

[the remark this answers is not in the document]
ananth102: With regards to the formatting, I think its just because of github.

pkg/resource/notebook_instance/hooks.go
ananth102: done

test/e2e/tests/test_notebook_instance.py
ananth102: done

templates/notebook_instance/sdk_update_pre_build_request.go.tpl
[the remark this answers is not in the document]
surajkota: my bad, you are right

surajkota (approved):
Good job!

/approve

ryansteakley (approved):

akartsky (approved):

mbaijal (approved):

PR #66, opened 2021-07-29
Unit Testing Framework w/ Feature Group sample

BriannaRoskind:
Summary:
Added unit testing framework (and code coverage), with an example of a unit test in feature group.

Code Coverage: 1.8%

Files Added:

- test/mocks/aws-sdk-go/sagemaker/SageMakerAPI.go - Contains unit testing api mocks.
- pkg/resource/feature_group/manager_test_suite_test.go - Contains resource specific helper functions for unit testing, and the provideResourceManagerWithMockSDKAPI function for unit testing.
- pkg/resource/feature_group/testdata/feature_group/v1alpha1/fg_invalid_before_create.yaml - Input sample file for unit testing.
- pkg/resource/feature_group/testdata/feature_group/v1alpha1/fg_invalid_create_attempted.yaml - Output (expected) sample file for unit testing.
- pkg/resource/feature_group/testdata/test_suite.yaml - Contains unit testing scenarios (which input/output to test along with the functions (Create/ReadOne/Update/Delete) to test).
- pkg/testutil/test_suite_config.go - Defines structures related to the unit testing test suite.
- pkg/testutil/test_suite_runner.go - Contains main workflow functions for running unit tests.
- pkg/testutil/util.go - Contains helper functions used in test_suite_runner.go.
- test/scripts/install-mockery.sh - A script that installs the mockery CLI tool that is used to build Go mocks for our interfaces to use in unit testing.

Files Edited:

- Makefile - Defined unit testing, mock, and unit testing coverage related testing calls.

PR #68, opened 2021-07-29
[Feature Group] Unit Testing

BriannaRoskind:
Note:
This PR is meant to be merged on top of [PR #66](https://github.com/aws-controllers-k8s/sagemaker-controller/pull/66).

Summary:
Added unit test to feature group for Create, ReadOne, and Delete (Feature Group does not support Update). Includes testing of state transitions. Code coverage for Feature Group = 53.9%.

Total Code Coverage: 3.1%

Files Added:

- pkg/resource/feature_group/testdata/feature_group/v1alpha1 - Contains yaml files to be used in the desired and expected fields for testing scenarios in test_suite.yaml. (8 files)
   - fg_before_create.yaml
   - fg_create_failed_state.yaml
   - fg_create_initiated.yaml
   - fg_created_state.yaml
   - fg_creating_state.yaml
   - fg_deleting_state.yaml
   - fg_invalid_before_create.yaml
   - fg_invalid_create_attempted.yaml
- pkg/resource/feature_group/testdata/feature_group/create - Contains json files to be used in the output_fixture fields for testing scenarios in test_suite.yaml which invoke the Create operation. (1 file)
   - fg_before_create.json
- pkg/resource/feature_group/testdata/feature_group/read_one- Contains json files to be used in the output_fixture fields for testing scenarios in test_suite.yaml which invoke the ReadOne operation. (4 files)
   - fg_create_failed.json
   - fg_created.json
   - fg_creating.json
   - fg_deleting.json

Files Edited:

- pkg/common/custom_conditions.go - fixed bug revealed by unit test (needed to remove the space in "Create Failed" to do a compare to the terminal state)
- pkg/resource/feature_group/manager_test_suite_test.go - Added YamlEqual that is used in test_suite_runner.go and compares the expected and actual resource as yaml files, presenting a diff output (and a copy of the actual output) which is easy to understand and use to modify the yaml files for testing. Also changed the Equal function to ignore the "LastTransitionTime" since as a timestamp it is updated while testing, and cannot be correctly preset in the expected output yaml file.
- pkg/resource/feature_group/testdata/test_suite.yaml - Added state transition testing.
- pkg/testutil/test_suite_runner.go - Added assert for the YamlEqual function to test the outputs of unit testing, and added YamlEqual to the TestRunnerDelegate interface. Also added option for unit testing scenario sdkapi mocking to have no provided output_fixture or error, and therefore mock both as being nil.

### DIFF
--- a/pkg/resource/feature_group/manager_test_suite_test.go
+++ b/pkg/resource/feature_group/manager_test_suite_test.go
@@ -16,6 +16,12 @@ package feature_group
 import (
 	"errors"
 	"fmt"
+	"os"
+	"io/ioutil"
+	"os/exec"
+
+	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
+	"github.com/ghodss/yaml"
 	mocksvcsdkapi "github.com/aws-controllers-k8s/sagemaker-controller/test/mocks/aws-sdk-go/sagemaker"
 	"github.com/aws-controllers-k8s/sagemaker-controller/pkg/testutil"
 	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
@@ -27,6 +33,12 @@ import (
 	ctrlrtzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
 	ackmetrics "github.com/aws-controllers-k8s/runtime/pkg/metrics"
 	"go.uber.org/zap/zapcore"
+)
+
+var (
+	DefaultTimestamp = "0001-01-01T00:00:00Z"
+	ReplaceTimestampRegExp = "s/\"[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z\"/\"" + DefaultTimestamp + "\"/"
+	TestDataDirectory = "testdata"
 )
 
 // provideResourceManagerWithMockSDKAPI accepts MockSageMakerAPI and returns pointer to resourceManager
@@ -48,8 +60,8 @@ func provideResourceManagerWithMockSDKAPI(mockSageMakerAPI *mocksvcsdkapi.SageMa
 	}
 }
 
-// TestDeclarativeTestSuite runs the test suite for feature group
-func TestDeclarativeTestSuite(t *testing.T) {
+// TestFeatureGroupTestSuite runs the test suite for feature group
+func TestFeatureGroupTestSuite(t *testing.T) {
      	defer func() {
      	   if r := recover(); r != nil {
      	      fmt.Println(testutil.RecoverPanicString, r)
@@ -88,7 +100,7 @@ func (d *testRunnerDelegate) EmptyServiceAPIOutput(apiName string) (interface{},
 	switch apiName {
 	case "CreateFeatureGroupWithContext":
 		var output svcsdk.CreateFeatureGroupOutput
-	    	return &output, nil
+		return &output, nil
 	case "DeleteFeatureGroupWithContext":
 		var output svcsdk.DeleteFeatureGroupOutput
 		return &output, nil
@@ -102,7 +114,8 @@ func (d *testRunnerDelegate) EmptyServiceAPIOutput(apiName string) (interface{},
 func (d *testRunnerDelegate) Equal(a acktypes.AWSResource, b acktypes.AWSResource) bool {
 	ac := a.(*resource)
 	bc := b.(*resource)
-	opts := []cmp.Option{cmpopts.EquateEmpty()}
+	// Ignore LastTransitionTime since it gets updated each run.
+	opts := []cmp.Option{cmpopts.EquateEmpty(), cmpopts.IgnoreFields(ackv1alpha1.Condition{}, "LastTransitionTime")}
 
 	if cmp.Equal(ac.ko.Status, bc.ko.Status, opts...) {
 		return true
@@ -111,4 +124,79 @@ func (d *testRunnerDelegate) Equal(a acktypes.AWSResource, b acktypes.AWSResourc
 		fmt.Println(cmp.Diff(ac.ko.Status, bc.ko.Status, opts...))
 		return false
 	}
+}
+
+// Checks to see if the given yaml file, with name stored as expectation,
+// matches the yaml marshal of the AWSResource stored as actual.
+func (d *testRunnerDelegate) YamlEqual(expectation string, actual acktypes.AWSResource) bool {
+     	// Get the file name of the expected yaml.
+	expectedYamlFileName := TestDataDirectory + "/" + expectation
+
+	// Build a tmp file for the actual yaml.
+	actualResource := actual.(*resource)
+	actualYamlByteArray, _ := yaml.Marshal(actualResource.ko)
+	actualYamlFileName := buildTmpFile("actualYaml", actualYamlByteArray)
+	defer os.Remove(actualYamlFileName)
+	if "" == actualYamlFileName {
+	  fmt.Printf("Could not create temporary actual file.\n")
+	  return false
+	}
+
+	// Replace Timestamps that would show up as different.
+	_, err := exec.Command("sed", "-r", "-i", ReplaceTimestampRegExp, actualYamlFileName).Output()
+	if isExecCommandError(err) {
+	    return false
+	}
+	
+	output,err := exec.Command("diff", "-c", expectedYamlFileName, actualYamlFileName).Output()
+	if isExecCommandError(err) {
+	   return false
+	}
+
+	if len(output) > 0 {
+	   actualOutput,err := exec.Command("cat", actualYamlFileName).Output()
+	   if isExecCommandError(err) {
+	      return false
+	   }
+	   fmt.Printf("\nExpected Yaml File Name: " + expectedYamlFileName + "\n")
+	   fmt.Printf("\nActual Output Yaml:\n" + string(actualOutput) + "\n")
+	   fmt.Printf("Diff From Expected:\n" + string(output) + "\n")
+	   return false
+	}
+	return true
+}
+
+func buildTmpFile(fileNameBase string, contents []byte) string {
+     newTmpFile, err := ioutil.TempFile(TestDataDirectory, fileNameBase)
+     if err != nil {
+        fmt.Println(err)
+	return ""
+     }
+     if _, err := newTmpFile.Write(contents); err != nil {
+        fmt.Println(err)
+	return ""
+     }
+     if err := newTmpFile.Close(); err != nil {
+        fmt.Println(err)
+	return ""
+     }		
+     return newTmpFile.Name()
+}
+
+// isExecCommandError returns true if an error
+// that is not an ExitError is found.
+func isExecCommandError(err error) bool {
+     if err == nil {
+     	return false
+     }
+     switch err.(type) {
+	case *exec.ExitError:
+	       // ExitError is expected.
+	       return false
+	default:
+	       // Couldn't run diff.
+	       fmt.Printf("Exec Command Error: ")
+	       fmt.Println(err)
+	       return true
+     }
 }
--- a/pkg/resource/feature_group/testdata/test_suite.yaml
+++ b/pkg/resource/feature_group/testdata/test_suite.yaml
@@ -2,16 +2,89 @@ tests:
   - name: "Feature group create test."
     description: "Feature group CRD tests"
     scenarios:
-     - name: "Create=InvalidInput"
-       description: "Given one of the parameters is invalid, ko.Status shows a terminal condition"
-       given:
-         desired_state: "feature_group/v1alpha1/fg_invalid_before_create.yaml"
-         svc_api:
-           - operation: CreateFeatureGroupWithContext
-             error:
-               code: InvalidParameterValue
-               message: "The feature group name must start with an alphanumeric character."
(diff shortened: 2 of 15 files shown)


pkg/common/custom_conditions.go
```diff
@@ -56,7 +57,7 @@ func SetTerminalState(
 	resourceName *string,
 	terminalStatus string,
 ) bool {
-	if latestStatus == nil || *latestStatus != terminalStatus {
+	if latestStatus == nil || strings.ReplaceAll(*latestStatus, " ", "") != terminalStatus {
```
akartsky, line 60: Oh nice unit tests caught a bug already

------------------------------------------------

why does`latestStatus` have a spaces in the first place ?

shouldn't `latestStatus` should already be set to one of these values `'Creating'|'Created'|'CreateFailed'|'Deleting'|'DeleteFailed'` ?

pkg/resource/feature_group/manager_test_suite_test.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	
+	output,err := exec.Command("diff", "-c", expectedYamlFileName, actualYamlFileName).Output()
+	if !checkExecCommandError(err) {
+	   return false
+	}
+
+	if len(output) > 0 {
+	   actualOutput,_ := exec.Command("cat", actualYamlFileName).Output()
+	   fmt.Printf("\nActual Yaml File Instead of: " + expectedYamlFileName + "\n" + string(actualOutput) + "\n")
```
akartsky, line 158: i feel wording is kinda weird 

Can we just print : 
Expected Yaml File Name :  `Full path to file expectedYamlFileName` 
Actual Output Yaml : `string(actualOutput)` 
Diff : `string(output)`

pkg/resource/feature_group/manager_test_suite_test.go
```diff
@@ -0,0 +1,196 @@
...
+	} else {
+		fmt.Printf("Difference (-expected +actual):\n\n")
+		fmt.Println(cmp.Diff(ac.ko.Status, bc.ko.Status, opts...))
+		return false
+	}
+}
+
+// Checks to see if the given yaml file, with name stored as expectation,
+// matches the yaml marshal of the AWSResource stored as actual.
+func (d *testRunnerDelegate) YamlEqual(expectation string, actual acktypes.AWSResource) bool {
```
akartsky, line 131: Nice

pkg/resource/feature_group/manager_test_suite_test.go
```diff
@@ -0,0 +1,196 @@
...
+	actualYamlByteArray, _ := yaml.Marshal(actualResource.ko)
+	actualYamlFileName := buildTmpFile("actualYaml", actualYamlByteArray)
+	defer os.Remove(actualYamlFileName)
+	if "" == actualYamlFileName {
+	  fmt.Printf("Could not create temporary actual file.\n")
+	  return false
+	}
+
+	// Replace Timestamps that would show up as different.
+	_, err := exec.Command("sed", "-r", "-i", ReplaceTimestampRegExp, actualYamlFileName).Output()
```
akartsky, line 146: Nit: 
I feel like it would be better to just do string manipulation before creating the temp file on line 138

pkg/testutil/test_suite_runner.go
```diff
@@ -0,0 +1,202 @@
...
+	latest acktypes.AWSResource
+	err    error
+}
+
+// TestRunnerDelegate provides interface for custom resource tests to implement.
+// TestSuiteRunner depends on it to run tests for custom resource.
+type TestRunnerDelegate interface {
+	ResourceDescriptor() acktypes.AWSResourceDescriptor
+	Equal(desired acktypes.AWSResource, latest acktypes.AWSResource) bool // remove it when ResourceDescriptor.Delta() is available
+	YamlEqual (expected string, actual acktypes.AWSResource) bool // new
```
akartsky, line 55: Nit : 
remove the space between YamlEqual and the open bracket

pkg/testutil/test_suite_runner.go
```diff
@@ -0,0 +1,202 @@
...
+		delta := runner.Delegate.ResourceDescriptor().Delta(expectedLatest, actual)
+		assert.Equal(0, len(delta.Differences))
+		if len(delta.Differences) > 0 {
+			fmt.Println("Unexpected differences:")
+			for _, difference := range delta.Differences {
+				fmt.Printf("Path: %v, expected: %v, actual: %v\n", difference.Path, difference.A, difference.B)
+			}
+		}
+
+		// Check that the yaml files are equivalent.
```
akartsky, line 138: can you also add the reason why we are doing this because one might get confused as to why we need it apart from `delta` and `Delegate.Equal`

Something like `// Check that the yaml files are equivalent. This makes it easier to make changes to unit test cases`

Also can you move this line to the end below the Delegate.Equal

pkg/resource/feature_group/testdata/test_suite.yaml
```diff
@@ -0,0 +1,87 @@
+tests:
+  - name: "Feature group demo test"
+    description: "Feature group CRD tests"
+    scenarios:
+      - name: "Create=InvalidInput"
```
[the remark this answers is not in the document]
BriannaRoskind: The naming convention that elasticache framework (which we have adopted) was using was to have the name of the operation (Create; ReadOne; Update; Delete) followed by a small description of how the test should behave/what is being tested (in this case InvalidInput).

pkg/common/custom_conditions.go
BriannaRoskind: This function was called from pkg/resource/feature_group/custom_update_conditions.go:

The latest status is retrieved by: r.ko.Status.FeatureGroupStatus.
When I printed this (for the CreateFailed state), it had a space in it ("Create Failed").
Whereas the terminal condition is retrieved by: svcsdk.FeatureGroupStatusCreateFailed (which has no spaces "CreateFailed")

pkg/testutil/test_suite_runner.go
BriannaRoskind: I made the comment change you mentioned.

However I kept the YamlEqual call before the Equal call because if the Equal call is first and the Equal assert fails (something in the status doesn't match) then the test stops/fails right there and you are not able to see the YamlEqual output which is more useful for debugging.

pkg/resource/feature_group/manager_test_suite_test.go
BriannaRoskind: Given that we are doing a line by line diff as an exec.Command, the sed matches very nicely with that intent which is line by line comparison, and appears to me to be clearer to the reader.

It would also be more code and a little less clear to iterate through the fields of the yaml object (checking their types).
Unless is there a really easy way to iterate through a yaml object checking the type of each field and then doing a replace.

pkg/common/custom_conditions.go
BriannaRoskind: This was caused by a typo in unit testing and has since been corrected. (Status was listed as "Create Failed" with a space rather than the expected "CreateFailed")

pkg/resource/feature_group/custom_update_api_test.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	mocksvcsdkapi "github.com/aws-controllers-k8s/sagemaker-controller/mocks/aws-sdk-go/sagemaker"
+	ctrlrtzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
+	ackmetrics "github.com/aws-controllers-k8s/runtime/pkg/metrics"
+	"go.uber.org/zap/zapcore"
+)
+
+// provideResourceManagerWithMockSDKAPI accepts MockSageMakerAPI and returns pointer to resourceManager
+// the returned resourceManager is configured to use mockapi api.
+func provideResourceManagerWithMockSDKAPI(mockSageMakerAPI *mocksvcsdkapi.SageMakerAPI) *resourceManager {
```
surajkota, line 26: why is this file named `custom_update_api_test`

this was part of manager_test_suite.go in other PR

pkg/resource/feature_group/manager_test_suite_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/aws-controllers-k8s/sagemaker-controller/pkg/testutil"
+	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
+	svcsdk "github.com/aws/aws-sdk-go/service/sagemaker"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"path/filepath"
+	"testing"
+)
+
+var	DefaultTimestamp = "0001-01-01T00:00:00Z"
```
surajkota, line 35: nit:
var (
)

akartsky (approved):

pkg/resource/feature_group/manager_test_suite_test.go
surajkota: ++

pkg/resource/feature_group/manager_test_suite_test.go
```diff
@@ -0,0 +1,167 @@
...
+	return ""
+     }
+     if err := newTmpFile.Close(); err != nil {
+        fmt.Println(err)
+	return ""
+     }		
+     return newTmpFile.Name()
+}
+
+func checkExecCommandError(err error) bool {
```
surajkota, line 153: nit: `isExecCmdError` might be more readable and return true when there is an actual error instead of false

pkg/resource/feature_group/testdata/feature_group/cr/fg_cmd_create_initiated.yaml
```diff
@@ -0,0 +1,23 @@
...
+  offlineStoreConfig:
+    s3StorageConfig:
+      s3URI: s3://source-data-bucket-592697580195-us-west-2/sagemaker/feature-group-data
+  recordIdentifierFeatureName: TransactionID
+  roleARN: arn:aws:iam::123456789012:role/ack-sagemaker-execution-role
+status:
+  ackResourceMetadata:
+    arn: arn:aws:sagemaker:us-west-2:123456789012:feature-group/unit-testing-feature-group
+    ownerAccountID: ""
+  conditions: []
```
surajkota, line 23: shouldn't this have a featureGroupStatus `Creating` as well?

pkg/resource/feature_group/testdata/feature_group/cr/fg_cmd_invalid_before_create.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: unit-testing-feature-group
+spec:
+  eventTimeFeatureName: EventTime
+  featureDefinitions:
+  - featureName: TransactionID
+    featureType: Integral
+  - featureName: EventTime
+    featureType: Fractional
+  featureGroupName: !unit-testing-feature-group
```
surajkota, line 12: nit: add a comment above this. invalid name or change the name to say so

pkg/resource/feature_group/testdata/test_suite.yaml
```diff
@@ -0,0 +1,87 @@
...
+          desired_state: "feature_group/cr/fg_cmd_creating_state.yaml" # Feature group is creating, but creating has not yet finished
+          svc_api:
+            - operation: DescribeFeatureGroupWithContext
+              output_fixture: "feature_group/read_one/fg_cmd_create_failed.json"
+        invoke: ReadOne
+        expect:
+          latest_state: "feature_group/cr/fg_cmd_create_failed_state.yaml"
+          error: nil
+      - name: "ReadOne=Created"
+        description: "A feature group in the creating state which enters the created state will have the ACK>ResourceSynced condition set to true, and the offlineStoreConfig dataCatalogConfig should be updated in the spec."
```
surajkota, line 52: `ACK.ResourceSynced`

pkg/resource/feature_group/testdata/test_suite.yaml
```diff
@@ -0,0 +1,87 @@
...
+        given:
+          desired_state: "feature_group/cr/fg_cmd_creating_state.yaml" # feature group is creating, but creating has not yet finished
```
surajkota, line 54: nit: write comment above the line

pkg/resource/feature_group/testdata/feature_group/cr/fg_cmd_invalid_before_create.yaml
BriannaRoskind: This name was changed in PRs #66 and #76. It has now been added to this one as well.

pkg/resource/feature_group/testdata/feature_group/cr/fg_cmd_create_initiated.yaml
BriannaRoskind: When Creating a feature group, it first calls create, which verifies all the input fields are of an acceptable format, as well as assigns a resource arn. After this, a ReadOne is called, where the status, conditions, and any additional fields (added through the custom_delta) are added.

pkg/resource/feature_group/custom_update_api_test.go
BriannaRoskind: This change was added in PRs #66 and #76. It has now been added to this one as well.

surajkota:
/approve

lmk once #66 merges and you rebase this. I can lgtm on this

PR #85, opened 2021-08-26
Use tmp file in unit-tests to allow for testing portability

ryansteakley:
Description of changes:
Unit tests fail when not using GNU sed this is because command uses `sed -i` which when not using GNU sed needs to specify a string afterwards which can be defaulted to `sed -i ''` However when attempting this allows for the unit-tests to pass but it created tmp files on my end with suffix ''. Ended up creating .tmp files initially and then removing them, this is consistent with the GNU internal implementation. https://www.gnu.org/software/sed/manual/sed.html#index-_002di 

> GNU sed does this by creating a temporary file and sending output to this file rather than to the standard output

### DIFF
--- a/pkg/testutil/util.go
+++ b/pkg/testutil/util.go
@@ -88,10 +88,7 @@ func IsYamlEqual(expectation *string, actualYamlByteArray *[]byte) bool {
 	// Remove tmp files used as backup https://riptutorial.com/sed/topic/9436/bsd-macos-sed-vs--gnu-sed-vs--the-posix-sed-specification
 	// -i inplace-editing is not consistent on both GNU and non-GNU sed when not specifiying a backup file.
 	actualYamlFileNameTmp := actualYamlFileName + ".tmp"
-	_, err = exec.Command("rm", actualYamlFileNameTmp).Output()
-	if isExecCommandError(err) {
-		return false
-	}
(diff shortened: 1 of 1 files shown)


pkg/testutil/util.go
```diff
@@ -79,7 +80,15 @@ func IsYamlEqual(expectation *string, actualYamlByteArray *[]byte) bool {
...
-	_, err := exec.Command("sed", "-r", "-i", ReplaceTimestampRegExp, actualYamlFileName).Output()
+	_, err := exec.Command("sed", "-r", "-i.tmp", ReplaceTimestampRegExp, actualYamlFileName).Output()
+	if isExecCommandError(err) {
+		return false
+	}
+
+	// Remove tmp files used as backup https://riptutorial.com/sed/topic/9436/bsd-macos-sed-vs--gnu-sed-vs--the-posix-sed-specification
+	// -i inplace-editing is not consistent on both GNU and non-GNU sed when not specifiying a backup file.
+	actualYamlFileNameTmp := actualYamlFileName + ".tmp"
+	_, err = exec.Command("rm", actualYamlFileNameTmp).Output()
```
surajkota, line 91: defer os.remove might work

akartsky (approved):

surajkota (approved):
/approve
/lgtm

PR #86, opened 2021-08-27
Update to runtime v0.13.0 

ryansteakley:
Description of changes:

- Updates to runtime v0.13.0. Fixes all issues with test-failures occurring as a result of updating the runtime.
- Fixes Notebook instance update, so that status is correctly updated.
- Utilizes `late_initializer` for new diffs that occur as a result of using `reflect.deepequal()`.
- Adds `new custom_delta.go` code for diffs that cannot be handled by `late_initializer`.


pkg/resource/training_job/custom_delta.go
```diff
@@ -46,4 +46,34 @@ func customSetDefaults(
 			a.ko.Spec.OutputDataConfig.KMSKeyID = b.ko.Spec.OutputDataConfig.KMSKeyID
 		}
 	}
+
+	// Default value of VolumeSizeInGB is 0
+	if ackcompare.IsNotNil(a.ko.Spec.ProfilerRuleConfigurations) && ackcompare.IsNotNil(b.ko.Spec.ProfilerRuleConfigurations) {
+		for index := range a.ko.Spec.ProfilerRuleConfigurations {
+			if ackcompare.IsNil(a.ko.Spec.ProfilerRuleConfigurations[index].VolumeSizeInGB) && ackcompare.IsNotNil(b.ko.Spec.ProfilerRuleConfigurations[index].VolumeSizeInGB) {
```
A-Hilaly, line 53: I'm not sure if this is completely correct. What if the arrays have different sizes?

A-Hilaly: Also i don't think you'll ever see a `nil` element in an array.

pkg/resource/processing_job/custom_delta.go
```diff
@@ -32,4 +32,31 @@ func customSetDefaults(
...
+			if ackcompare.IsNil(a.ko.Spec.ProcessingInputs[index].AppManaged) && ackcompare.IsNotNil(b.ko.Spec.ProcessingInputs[index].AppManaged) {
+				a.ko.Spec.ProcessingInputs[index].AppManaged =
+					b.ko.Spec.ProcessingInputs[index].AppManaged
+			}
+			if ackcompare.IsNil(a.ko.Spec.ProcessingInputs[index].S3Input.S3DataDistributionType) && ackcompare.IsNotNil(b.ko.Spec.ProcessingInputs[index].S3Input.S3DataDistributionType) {
+				a.ko.Spec.ProcessingInputs[index].S3Input.S3DataDistributionType =
+					b.ko.Spec.ProcessingInputs[index].S3Input.S3DataDistributionType
+			}
+		}
+	}
```
A-Hilaly, line 49: ditto

pkg/resource/notebook_instance/hooks.go
```diff
@@ -66,6 +66,10 @@ func (rm *resourceManager) customSetOutputDescribe(r *resource) error {
 				return err
 			}
 			r.ko.Status.StoppedByControllerMetadata = nil
+			// Notebook after rm.startNotebookInstance(r) is in pending state
+			// Manually change status since otherwise synced condition will be set to True as
+			// of a result Stopped status.
+			r.ko.Status.NotebookInstanceStatus = aws.String(svcsdk.NotebookInstanceStatusPending)
```
A-Hilaly, line 72: Is this correct whenever the reconciler is doing a ReadOne call?

pkg/resource/training_job/custom_delta.go
ryansteakley: the nil check on line 53 is in reference to VolumeSizeInGB which is an element of a map. As its an array of maps/dicts.

pkg/resource/notebook_instance/hooks.go
ryansteakley: The reconciler is doing a readOne call however it will get the latest status as Stopped which will cause the reconciler to no longer reconcile, so we will not be able to see the ensuing pending/in-service status that occur after the noteBook instance is started after being in the stopped state during this update codepath

pkg/testutil/util.go
```diff
@@ -79,11 +80,16 @@ func IsYamlEqual(expectation *string, actualYamlByteArray *[]byte) bool {
 	}
 
 	// Replace Timestamps that would show up as different.
-	_, err := exec.Command("sed", "-r", "-i", ReplaceTimestampRegExp, actualYamlFileName).Output()
+	_, err := exec.Command("sed", "-r", "-i.tmp", ReplaceTimestampRegExp, actualYamlFileName).Output()
```
RedbackThomson, line 83: Why are we using command line arguments within Golang to begin with? Go has support for marshaling and unmarshaling YAML into structs, and Go supports determining differences between structs. Especially when deleting or reading files (the `cat` lower down), there is no reason NOT to use the `os` package.

surajkota (changes requested):

pkg/testutil/test_suite_runner.go
```diff
@@ -17,13 +17,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"path/filepath"
```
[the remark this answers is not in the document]
ryansteakley: Whenever I save it automatically appears on my end

pkg/resource/notebook_instance_lifecycle_config/delta.go
```diff
@@ -39,6 +48,12 @@ func newResourceDelta(
...
 		}
 	}
+	if !reflect.DeepEqual(a.ko.Spec.OnCreate, b.ko.Spec.OnCreate) {
+		delta.Add("Spec.OnCreate", a.ko.Spec.OnCreate, b.ko.Spec.OnCreate)
+	}
+	if !reflect.DeepEqual(a.ko.Spec.OnStart, b.ko.Spec.OnStart) {
+		delta.Add("Spec.OnStart", a.ko.Spec.OnStart, b.ko.Spec.OnStart)
+	}
 
 	delta = customDeltaOnCreate(delta, a, b)
```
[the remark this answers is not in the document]
ryansteakley: Not sure, I can test it without custom implementation

pkg/testutil/util.go
ryansteakley: This is just a temp fix to get unit tests working on MACOSX, the unit test code here can be rewritten in a separate pr.

pkg/resource/training_job/custom_delta.go
```diff
@@ -46,4 +47,36 @@ func customSetDefaults(
...
+				a.ko.Spec.ProfilerRuleConfigurations[index].VolumeSizeInGB = DefaultVolumeSizeInGB
+			}
+		}
+	}
+
+	// Default value of VolumeSizeInGB is 0
+	if ackcompare.IsNotNil(a.ko.Spec.DebugRuleConfigurations) {
+		for index := range a.ko.Spec.DebugRuleConfigurations {
+			if ackcompare.IsNil(a.ko.Spec.DebugRuleConfigurations[index].VolumeSizeInGB) {
+
```
surajkota, line 66: nit: extra line

pkg/resource/training_job/custom_delta.go
```diff
@@ -46,4 +47,36 @@ func customSetDefaults(
 			a.ko.Spec.OutputDataConfig.KMSKeyID = b.ko.Spec.OutputDataConfig.KMSKeyID
 		}
 	}
+
+	// Default value of VolumeSizeInGB is 0
+	DefaultVolumeSizeInGB := aws.Int64(0)
```
surajkota, line 52: nit: `defaultVolumeSizeInGB`

use camel case for variable names here and other files too

pkg/resource/model_package/testdata/model_package/v1alpha1/mp_invalid_before_create.yaml
```diff
@@ -29,6 +29,7 @@ spec:
                 s3URI: s3://source-data-bucket-592697580195-us-west-2/sagemaker/batch/input-data
             contentType: text/csv
           transformOutput:
+            kmsKeyID: ""
```
surajkota, line 32: desired may or may not specify these values, lets add unit tests separately and leave them out of this PR

and since the create failed, `mp_invalid_create_attempted.yaml` will also not have this value, since its coming as response of describe call

we will have to test this out when we have a readone or update related test

same for other unit test changes in this PR

pkg/resource/notebook_instance_lifecycle_config/delta.go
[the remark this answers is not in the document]
surajkota: true, but we don't need both. I guess we can ignore these fields for generic implementation and rely on custom implementation since that is robust

surajkota (changes requested):
Few comments and some of the earlier comments need to be addressed

apis/v1alpha1/generator.yaml
```diff
@@ -673,3 +743,4 @@ ignore:
       - Workteam
     operations:
       - UpdateTrainingJob
+
```
surajkota, line 746: nit: extra line

surajkota (approved):
/approve
/lgtm

Nice work on this PR. small nit fix if tests need to be rerun

Note: reflect.deepEqual does an order-dependent comparison. The code in this PR is written assuming the service side response retains the order in which the request was sent. This is not the best solution but is a good stop-gap until we have a better solution and maybe option to specify algorithms for compare

For service side defaults, we considered not looking at b/server response and just assigning the default but it came with a con that those fields appeared like required in unit tests because we run [delta separately]( https://github.com/aws-controllers-k8s/sagemaker-controller/blob/main/pkg/testutil/test_suite_runner.go#L142), so we still check for isNotNil(b)

apis/v1alpha1/generator.yaml
```diff
@@ -38,6 +38,10 @@ resources:
         - InvalidParameterCombination
         - InvalidParameterValue
         - MissingParameter
+    fields:
+      Tags:
```
mbaijal, line 42: just confirming so I can pick this change up - is this as simple as this field wont be considered by the diff checker ?

helm/crds/sagemaker.services.k8s.aws_trainingjobs.yaml
```diff
@@ -516,6 +516,32 @@ spec:
...
                 description: If the training job failed, the reason it failed.
                 type: string
+              modelArtifacts:
+                description: Information about the Amazon S3 location that is configured
+                  for storing model artifacts.
+                properties:
+                  s3ModelArtifacts:
+                    type: string
+                type: object
+              profilerRuleEvaluationStatuses:
```
mbaijal, line 526: why is this a part of this PR, SDK version hasn't changed

helm/values.yaml
```diff
@@ -37,17 +37,21 @@ aws:
   # If specified, use the AWS region for AWS API calls
   region: ""
   account_id: ""
+  endpoint_url: ""
 
 # log level for the controller
 log:
-  enable_development_logging: true
-  level: debug
+  enable_development_logging: false
```
mbaijal, line 44: We need to also revert the tag back to 0.0.3 in this file

pkg/resource/hyper_parameter_tuning_job/testdata/hyper_parameter_tuning_job/v1alpha1/hptj_invalid_before_create.yaml
```diff
@@ -18,8 +18,6 @@ spec:
         minValue: '10'
         maxValue: '20'
         scalingType: Linear
-      continuousParameterRanges: []
```
mbaijal, line 21: what changed to require this change to testdata or was it redundant to begin with ?

ryansteakley: Delta now using deepReflect so having this causes a diff which thus causes the unit-test to fail.

helm/crds/sagemaker.services.k8s.aws_trainingjobs.yaml
ryansteakley: Was auto-generated via code-gen

apis/v1alpha1/generator.yaml
ryansteakley: Correct delta.go will not check for diffs in tags

pkg/resource/hyper_parameter_tuning_job/testdata/hyper_parameter_tuning_job/v1alpha1/hptj_invalid_before_create.yaml
mbaijal: but we should be handling the diff generated correct instead of changing the unit test ?

helm/crds/sagemaker.services.k8s.aws_trainingjobs.yaml
mbaijal: hmm, not sure why it is on the PR but it should be there  in the helm either way, so ok.

surajkota: I was using an older version of code gen, this change is in the latest codegen

pkg/resource/hyper_parameter_tuning_job/testdata/hyper_parameter_tuning_job/v1alpha1/hptj_invalid_before_create.yaml
surajkota: In general, yes, but this is anyway invalid input so it's fine

ryansteakley: test_suite_runner.go:144: 
        	Error Trace:	test_suite_runner.go:144
        	            				test_suite_runner.go:115
        	            				test_suite_runner.go:91
        	Error:      	Not equal: 
        	            	expected: 0
        	            	actual  : 2
        	Test:       	TestHyperParameterTuningJobTestSuite/Create=InvalidInput
  ```
Unexpected differences:
  Path: {[Spec HyperParameterTuningJobConfig ParameterRanges CategoricalParameterRanges]}, expected: [], actual: []
  See expected differences below:
  Path: {[Spec HyperParameterTuningJobConfig ParameterRanges ContinuousParameterRanges]}, expected: [], actual: []
```

ryansteakley: In the before create yaml it has these two fields which are supposed to be lists of dicts/maps however they are empty. In the after yaml they are not included in the spec so, I decided to remove it from the test, can attempt to try to debug, however diff message is confusing.

mbaijal: I see, this is fine for now. Better to have a cleaner unit test.

mbaijal (approved):

PR #87, opened 2021-08-30
Implement per region instance type config for canary and e2e tests

rrkharse:
Description of changes:

Currently, we are using one fixed instance type across all AWS regions in our endpoint and training job tests. 
However, certain regions do not support the currently specified instance type or require a limit increase to use that instance type.
Specifically, canary tests in the eu-west-3 and eu-north-1 regions are failing due to this issue. 

This pull request updates the testing resource config file `replacement_values.py` to pass in the correct instance type depending on region. 
Regions that did not experience this issue will continue to use the previous instance type via the new config to avoid breaking canaries/e2e testing in those regions.

The changes have been tested in our eu-west-3 and eu-north-1 canary stacks and have resulted in passing canaries.

### DIFF
--- a/test/e2e/replacement_values.py
+++ b/test/e2e/replacement_values.py
@@ -167,49 +167,13 @@ CLARIFY_IMAGE_URIS = {
 }
 
 ENDPOINT_INSTANCE_TYPES = {
-    "us-east-1": "ml.c5.large",
-    "us-east-2": "ml.c5.large",
-    "us-west-1": "ml.c5.large",
-    "us-west-2": "ml.c5.large",
-    "ap-east-1": "ml.c5.large",
-    "ap-south-1": "ml.c5.large",
-    "ap-northeast-2": "ml.c5.large",
-    "ap-southeast-1": "ml.c5.large",
-    "ap-southeast-2": "ml.c5.large",
-    "ap-northeast-1": "ml.c5.large",
-    "ca-central-1": "ml.c5.large",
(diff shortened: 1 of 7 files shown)


test/e2e/replacement_values.py
```diff
@@ -175,4 +221,6 @@
     "SAGEMAKER_EXECUTION_ROLE_ARN": get_bootstrap_resources().ExecutionRoleARN,
     "MODEL_MONITOR_ANALYZER_IMAGE_URI": f"{MODEL_MONITOR_IMAGE_URIS[get_region()]}/sagemaker-model-monitor-analyzer",
     "CLARIFY_IMAGE_URI": f"{CLARIFY_IMAGE_URIS[get_region()]}/sagemaker-clarify-processing:1.0",
+    "ENDPOINT_INSTANCE_TYPE": ENDPOINT_INSTANCE_TYPES[get_region()],
```
surajkota, line 224: nit: since most regions, we can use same instance type, can we use
`ENDPOINT_INSTANCE_TYPES.get(get_region(), ml.c5.large)`

surajkota (changes requested):
one nit, otherwise looks good

test/e2e/replacement_values.py
rrkharse: Makes sense, added changes below

akartsky (approved):

surajkota (approved):
/approve
/lgtm

PR #88, opened 2021-08-30
requeue until Stopped for job style resources

surajkota:
Description of changes:
- requeue until Stopped for job style resources
  - integ test changes for the same
- Reduce terminal codes to only those which can be fixed by a change in spec. Reason being the latest runtime 0.13+ ignores metadata and status updates leaving customer with no way to get the controller reconcile if the terminal code is not related to spec field change. I will make this change for other resources in a separate PR
- consolidate hook related code to one file and refractor to use common pkg to reduce code duplication
  - removed post `sdk_create_post_set_output` hook because runtime does a readone and paches the status post create

TODO:
During testing, I found that if there is an error during delete it is not reflected in the resource conditions. Will put a PR to codegen and create a new PR with those changes


akartsky (approved):

test/e2e/tests/test_hpo.py
```diff
@@ -143,7 +143,7 @@ def test_stopped(self, xgboost_hpojob):
         )
 
         # Delete the k8s resource.
-        _, deleted = k8s.delete_custom_resource(reference, 3, 10)
+        _, deleted = k8s.delete_custom_resource(reference, cfg.JOB_DELETE_WAIT_PERIODS, cfg.JOB_DELETE_WAIT_PERIODS)
```
ryansteakley, line 146: I think you meant to use cfg.JOB_DELETE_WAIT_LENGTH here
```suggestion
        _, deleted = k8s.delete_custom_resource(reference, cfg.JOB_DELETE_WAIT_PERIODS, cfg.JOB_DELETE_WAIT_LENGTHS)
```

test/e2e/tests/test_hpo.py
```diff
@@ -56,7 +56,7 @@ def xgboost_hpojob():
     yield (reference, resource)
 
     if k8s.get_resource_exists(reference):
-        _, deleted = k8s.delete_custom_resource(reference, 3, 10)
+        _, deleted = k8s.delete_custom_resource(reference, cfg.JOB_DELETE_WAIT_PERIODS, cfg.JOB_DELETE_WAIT_PERIODS)
```
ryansteakley, line 59: I think you meant to use cfg.JOB_DELETE_WAIT_LENGTH here
```suggestion
        _, deleted = k8s.delete_custom_resource(reference, cfg.JOB_DELETE_WAIT_PERIODS, cfg.JOB_DELETE_WAIT_LENGTHS)
```

surajkota: thx, done

ryansteakley (approved):

PR #90, opened 2021-08-31
Enable adoption and versioned model package

ryansteakley:
Description of changes:

1. Re-enables versioned model-package as using updated runtime and deploying to canary did not see test failure for versioned model package. Updated runtime fixes issue of multiple reconciliation loops which appears to have caused multiple creations of model-package.
2. Latest runtime/codegen can now correctly adopt and retrieve status for versioned model-package, thus adoption is now enabled for model-package.


pkg/resource/model_package/resource.go
```diff
@@ -92,10 +92,15 @@ func (r *resource) SetStatus(desired acktypes.AWSResource) {
...
-	if identifier.NameOrID == "" {
-		return ackerrors.MissingNameIdentifier
+	if r.ko.Status.ACKResourceMetadata == nil {
+		r.ko.Status.ACKResourceMetadata = &ackv1alpha1.ResourceMetadata{}
+	}
+	r.ko.Status.ACKResourceMetadata.ARN = identifier.ARN
+
+	f0, f0ok := identifier.AdditionalKeys["modelPackageName"]
+	if f0ok {
+		r.ko.Spec.ModelPackageName = &f0
```
surajkota, line 102: shouldn't this be removed?

we just want to adopt using ARN for all cases right?

also, please add an integ test for this since its different from other resources

surajkota (changes requested):

pkg/resource/model_package/resource.go
ryansteakley: Atm looks like code-gen will add those lines of code regardless. I have set primary identifier to ARN. 
Added integ test for adopting versioned model package

surajkota: ok for now, let's come back to this

test/e2e/tests/test_adopt_model_package.py
```diff
@@ -0,0 +1,239 @@
...
+
+        model_package_reference = k8s.create_reference(
+            CRD_GROUP,
+            CRD_VERSION,
+            cfg.MODEL_PACKAGE_RESOURCE_PLURAL,
+            model_package_input["ModelPackageGroupName"] + "-child",
+            namespace,
+        )
+        model_package_resource = k8s.wait_resource_consumed_by_controller(
+            model_package_reference, wait_periods=30, period_length=30
```
surajkota, line 218: nit time

test/e2e/tests/test_adopt_model_package.py
```diff
@@ -0,0 +1,239 @@
...
+        assert k8s.get_resource_arn(
+            model_package_resource
+        ) == model_package_response.get("ModelPackageArn", None)
+
+        assert_model_package_status_in_sync(
+            model_package_arn, model_package_reference, cfg.JOB_STATUS_COMPLETED,
+        )
+        assert k8s.wait_on_condition(
+            model_package_reference, "ACK.ResourceSynced", "True"
+        )
```
surajkota, line 239: can add delete as well

surajkota (changes requested):

akartsky (approved):

surajkota (approved):
/approve

PR #91, opened 2021-09-01
update conditions on delete and aws error code+message

surajkota:
Description of changes:
- bring in changes from https://github.com/aws-controllers-k8s/code-generator/pull/175, https://github.com/aws-controllers-k8s/code-generator/pull/176
- corresponding unit and integ test changes
- fix some erroneous unit tests like model package group and endpoint config
- fix invalid name in unit test
- fix assertion in notebook instance test since PR build was failing and add it to missing canary cleanup

Testing:
unit test
tested manually, rely on PR build, later we can add more unit tests for delete failed scenario

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #92, opened 2021-09-01
remove status.EndpointConfig* from endpoint CRD

surajkota:
Description of changes:
1. Remove `status.latestEndpointConfgName` from CRD
  - This field was added for customers to have a reliable way to wait for a successful update because the endpoint goes back to `InService` even after a failed update and the `ResourceSynced` condition was set to `True`. 
  - A customer could not rely on `ResourceSynced` or `endpointStatus` to get the accurate picture and they needed do multiple checks the update is successful.
    -  e.g. `endpointStatus == InService && ACK.ResourceSynced == true && ACK.Terminal == False && failureReason = nil`.
    - To simplify this, latestEndpointConfig was added to the status so Cx can wait on just `spec.EndpointConfigName == status.latestEndpointConfigName`
  - With this change, the `ResourceSynced` condition will not go to `True` on failed update since the desired state cannot be reached. Customers can just wait and timeout on:
    - `endpointStatus == InService && ACK.ResourceSynced == true` and check for failure reason in case of timeout. The latestEndpointConfig from service will be in the Terminal condition message for convinence
2. Move `status.lastEndpointConfigForUpdate` to metadata
  - This is purely a metadata field. It was only required by the controller to stop retrying an update and hence has been moved to annotation.
  - Moving this to annotations gives us a two-way door. In case we find a better heuristic in future to differentiate between update related error and related to service modification, this can be removed easily with having any backward breaking changes to the CRD

Testing:
```
pytest tests/test_endpoint.py
```
PR build will pass after #91  is merged since there is a dependency on the error message

### DIFF
--- a/pkg/resource/endpoint/custom_update_conditions.go
+++ b/pkg/resource/endpoint/custom_update_conditions.go
@@ -17,13 +17,15 @@
 package endpoint
 
 import (
+	"github.com/aws/aws-sdk-go/aws"
+
 	ackcondition "github.com/aws-controllers-k8s/runtime/pkg/condition"
 	ackerr "github.com/aws-controllers-k8s/runtime/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	svcapitypes "github.com/aws-controllers-k8s/sagemaker-controller/apis/v1alpha1"
 	svccommon "github.com/aws-controllers-k8s/sagemaker-controller/pkg/common"
-	"github.com/aws/aws-sdk-go/aws"
 	svcsdk "github.com/aws/aws-sdk-go/service/sagemaker"
-	corev1 "k8s.io/api/core/v1"
+	
 )
 
 // CustomUpdateConditions sets conditions (terminal) on supplied endpoint.
@@ -46,7 +48,7 @@ func (rm *resourceManager) CustomUpdateConditions(
 	// since desired and latest will be different until the issue is fixed.
 	// Customer can use this condition state and FailureReason to determine
 	// the correct course of action in case the update to Endpoint fails
-	// Customer will also have additional information like latest endpointconfg
+	// Customer will also have additional information like latest endpointConfig
 	// in condition message and last endpointconfig used for update in annotations
 	if err != nil {
 		awsErr, ok := ackerr.AWSError(err)
--- a/pkg/resource/endpoint/custom_update_conditions.go
+++ b/pkg/resource/endpoint/custom_update_conditions.go
@@ -17,9 +17,13 @@
 package endpoint
 
 import (
+	ackcondition "github.com/aws-controllers-k8s/runtime/pkg/condition"
+	ackerr "github.com/aws-controllers-k8s/runtime/pkg/errors"
 	svcapitypes "github.com/aws-controllers-k8s/sagemaker-controller/apis/v1alpha1"
 	svccommon "github.com/aws-controllers-k8s/sagemaker-controller/pkg/common"
+	"github.com/aws/aws-sdk-go/aws"
 	svcsdk "github.com/aws/aws-sdk-go/service/sagemaker"
+	corev1 "k8s.io/api/core/v1"
 )
 
 // CustomUpdateConditions sets conditions (terminal) on supplied endpoint.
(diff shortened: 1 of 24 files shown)


goswamig (approved):

pkg/resource/endpoint/custom_update_conditions.go
```diff
@@ -36,5 +40,21 @@ func (rm *resourceManager) CustomUpdateConditions(
...
 	// If the latestStatus == terminalStatus we will set
 	// the terminal condition and terminal message.
-	return svccommon.SetTerminalState(conditionManager, latestStatus, &resourceName, terminalStatus)
+	updated := svccommon.SetTerminalState(conditionManager, latestStatus, &resourceName, terminalStatus)
+
+	// Continue setting ResourceSynced condition to false in case of failed update
+	// since desired and latest will be different until the issue is fixed.
+	// Customer can use this condition state and FailureReason to determine
+	// the correct course of action in case the update to Endpoint fails
+	// Customer will also have additional information like latest endpointconfg
```
ryansteakley, line 49: tiny nit: "endpointconfg" typo

pkg/resource/endpoint/hooks.go
```diff
@@ -98,7 +86,15 @@ func (rm *resourceManager) customUpdateEndpointPreChecks(
 
 	failureReason := latest.ko.Status.FailureReason
 	desiredEndpointConfig := desired.ko.Spec.EndpointConfigName
-	lastEndpointConfigForUpdate := desired.ko.Status.LastEndpointConfigNameForUpdate
+	
+	var lastEndpointConfigForUpdate *string = nil
+	// get last endpoint config name used for udapte form annotation
```
ryansteakley, line 91: nit: typo udapte form -> update from ?

test/e2e/tests/test_endpoint.py
```diff
@@ -33,8 +33,8 @@
 from e2e.replacement_values import REPLACEMENT_VALUES
 from e2e.common import config as cfg
 
-FAIL_UPDATE_ERROR_MESSAGE = "unable to update endpoint. check FailureReason"
-
+FAIL_UPDATE_ERROR_MESSAGE = "EndpointUpdateError: unable to update endpoint. check FailureReason. latest EndpointConfigName is "
+LAST_ENDPOINTCONFIGNNAME_FOR_UPDATE_ANNOTATION_KEY = "sagemaker.services.k8s.aws/Endpoint-last-endpoint-config-for-update"
```
akartsky, line 37: nit: lets keep the variable name length small  
you can write a comment above it to explain exactly what it is used for if necessary 

should be fine for now

akartsky (approved):

pkg/resource/endpoint/hooks.go
```diff
@@ -39,39 +37,29 @@ var (
 
 	resourceName = resourceGK.Kind
 
-	FailUpdateError = awserr.New("EndpointUpdateError", "unable to update endpoint. check FailureReason", nil)
+	lastEndpointConfigForUpdateAnnotation = fmt.Sprintf("%s/%s-last-endpoint-config-for-update", resourceGK.Group, resourceGK.Kind)
```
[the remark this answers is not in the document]
surajkota: thx

pkg/resource/endpoint/custom_update_conditions.go
```diff
@@ -17,9 +17,13 @@
 package endpoint
 
 import (
+	ackcondition "github.com/aws-controllers-k8s/runtime/pkg/condition"
+	ackerr "github.com/aws-controllers-k8s/runtime/pkg/errors"
 	svcapitypes "github.com/aws-controllers-k8s/sagemaker-controller/apis/v1alpha1"
 	svccommon "github.com/aws-controllers-k8s/sagemaker-controller/pkg/common"
+	"github.com/aws/aws-sdk-go/aws"
 	svcsdk "github.com/aws/aws-sdk-go/service/sagemaker"
+	corev1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
surajkota: thx, will look into in a later PR

PR #100, opened 2021-09-07
Use late-initialize where possible

ryansteakley:
Description of changes:
Use late-initialize code where it is applicable to replace custom-delta code with late-initialize in the generator.yaml

### DIFF
--- a/generator.yaml
+++ b/generator.yaml
@@ -43,6 +43,9 @@ resources:
       Tags:
         compare:
           is_ignored: true
+      EnableNetworkIsolation:
+        late_initialize:
+          min_backoff_seconds: 5
   EndpointConfig:
     exceptions:
       errors:
--- a/generator.yaml
+++ b/generator.yaml
@@ -179,6 +179,21 @@ resources:
       AlgorithmSpecification.MetricDefinitions:
         late_initialize:
           min_backoff_seconds: 5
+      EnableInterContainerTrafficEncryption:
+        late_initialize:
+          min_backoff_seconds: 5
+      EnableManagedSpotTraining:
+        late_initialize:
+          min_backoff_seconds: 5
+      EnableNetworkIsolation:
+        late_initialize:
+          min_backoff_seconds: 5
+      AlgorithmSpecification.EnableSageMakerMetricsTimeSeries:
+        late_initialize:
+          min_backoff_seconds: 5
+      OutputDataConfig.KMSKeyID:
+        late_initialize:
+          min_backoff_seconds: 5
       Tags:
         compare:
           is_ignored: true
@@ -218,7 +233,7 @@ resources:
           priority: 1
         from:
           operation: DescribeProcessingJob
-          path: FailureReason 
+          path: FailureReason      
       Tags:
         compare:
           is_ignored: true
@@ -259,6 +274,12 @@ resources:
         from:
           operation: DescribeTransformJob
           path: FailureReason
+      TransformInput.SplitType:
+        late_initialize:
+          min_backoff_seconds: 5   
+      TransformInput.CompressionType: 
+        late_initialize:
+          min_backoff_seconds: 5  
       Tags:
         compare:
           is_ignored: true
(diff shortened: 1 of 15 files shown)


generator.yaml
```diff
@@ -499,6 +532,8 @@ resources:
         from:
           operation: DescribeModelPackage
           path: ModelApprovalStatus
+        late_initialize:
```
surajkota, line 535: why do we need late initialize for this field?

surajkota (commented):
One question inline

2 fields I am doubtful about because they might fit under server-side overrides than server-side defaults. Basically need to answer, what happens if the customer specifies a part of input in these fields?

1. `AlgorithmSpecification.MetricDefinitions` in training job - e.g. is there a way to distinguish which metricDefinitions are added by Cx vs service similar to `TrainingJobDefinition.StaticHyperParameters` in HPO

2. `OfflineStoreConfig.S3StorageConfig.ResolvedOutputS3URI` in feature group. is there a pattern to this override? like what happens when a customer specifies an S3 path prefix, code in this PR won't be able to handle this since `a.spec....` is not nil

Can be a follow up PR though, dont need to block this PR to handle these cases

generator.yaml
ryansteakley: Refer to custom_delta for model package this field is not mandatory but is however set to a default value of 	`// Default is for ModelApprovalStatus to be set to pending manual approval`

pkg/resource/model/custom_delta.go
```diff
@@ -16,6 +16,7 @@ package model
 import (
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
 	svcapitypes "github.com/aws-controllers-k8s/sagemaker-controller/apis/v1alpha1"
+	"github.com/aws/aws-sdk-go/aws"
```
A-Hilaly, line 19: Please apply `goimport` here

pkg/resource/model/custom_delta.go
```diff
@@ -33,7 +34,20 @@ func customSetDefaults(
...
 	}
 
-	if ackcompare.IsNil(a.ko.Spec.EnableNetworkIsolation) && ackcompare.IsNotNil(b.ko.Spec.EnableNetworkIsolation) {
-		a.ko.Spec.EnableNetworkIsolation = b.ko.Spec.EnableNetworkIsolation
+	// Default value of Mode is SingleModel
+	mode := aws.String("SingleModel")
+
+	if ackcompare.IsNotNil(a.ko.Spec.Containers) && ackcompare.IsNotNil(b.ko.Spec.Containers) {
+		for index := range a.ko.Spec.Containers {
+			if ackcompare.IsNil(a.ko.Spec.Containers[index].Mode) && ackcompare.IsNotNil(b.ko.Spec.Containers[index].Mode) {
```
A-Hilaly, line 42: What if `len(a.ko.Spec.Containers) > len(b.ko.Spec.Containers)`?

A-Hilaly (commented):
`generator.yaml` looks good! 👍 
left two comments

mbaijal (approved):

pkg/resource/model/custom_delta.go
ryansteakley: Ideally we want to just set these server side defaults regardless however for our current unit-test implementation we need to check b, ideally we wouldn't. Can be improved on further pr

pkg/resource/model/custom_delta.go
ryansteakley: I downloaded and ran goimport but didn't seem to change anything? What is the expected output ?

generator.yaml
surajkota: let's remove this field from status since its part of the spec already

pkg/resource/model/custom_delta.go
surajkota: last I tried, it brought in a bunch of dependencies to the package.

mbaijal: same, but is there a package size concern or should we just go ahead and include it ?

PR #103, opened 2021-09-08
Remove model-approval-status read-only  from generator.yaml

ryansteakley:
Description of changes:
Remove model-approval-status as a read-only field it is not a status field, it is a part of the spec, don't need it in both places


generator.yaml
```diff
@@ -494,11 +494,6 @@ resources:
         from:
           operation: DescribeModelPackage
           path: ModelPackageStatusDetails
-      ModelApprovalStatus:
```
surajkota, line 497: something wrong here, it undoes the late initializer stuff?

ryansteakley: Opened pr before late_initialize got merged, will rebase

PR #114, opened 2021-09-28
Use available instance type for notebook-instance tests selected by region

ryansteakley:
- Use different instance type for `eu-north-1` for notebook_instance test
  - This is because the default instance type is not available on `eu-north-1`


mbaijal (approved):

mbaijal:
> * Use different instance type for `eu-north-1` for notebook_instance test

Can you add a reason here as to why this change was needed, would help anyone referring to the PR in the future.

akartsky (approved):

PR #117, opened 2021-10-05
Increase timeout for certain e2e status checks and re-order cleanup

ryansteakley:
Description of changes:

- Some Canary e2e tests are failing with status mismatch ex: `Inprogress != Completed`, even though the jobs will eventually pass.
- Reached out to service teams for investigation in what may be occurring on these jobs.  
- Some resources are order dependent for deletion ex: Cannot delete an `endpoint` if it has a `monitoringSchedule` applied to it.


surajkota (approved):
/approve
/lgtm

akartsky (approved):

PR #120, opened 2021-10-08
regen with codegen v0.15.1

ryansteakley:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/1014

Description of changes:
Update to codegen v0.15.1 and use runtime v0.15.1 in the runtime

- Uses `is_primary_arn` instead of `primary_identifier_field`
- `ResourceGK` was renamed to `GroupKind` so changed this field wherever we used it.

### DIFF
--- a/pkg/resource/endpoint/custom_update_conditions.go
+++ b/pkg/resource/endpoint/custom_update_conditions.go
@@ -38,7 +38,7 @@ func (rm *resourceManager) CustomUpdateConditions(
 	latestStatus := r.ko.Status.EndpointStatus
 	terminalStatus := svcsdk.EndpointStatusFailed
 	conditionManager := &resource{ko}
-	resourceName := resourceGK.Kind
+	resourceName := GroupKind.Kind
 	// If the latestStatus == terminalStatus we will set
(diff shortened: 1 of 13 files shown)


generator.yaml
```diff
@@ -13,18 +13,9 @@ operations:
       RetainAllVariantProperties: true
   DescribeModelPackage:
     custom_check_required_fields_missing_method: customCheckRequiredFieldsMissingMethod
-    primary_identifier_field_name: ARN
```
surajkota, line 16: I don't see this override in the new format. can you elaborate on how it works? because I see that model_package is_identifier has not changed

ryansteakley: Hey Suraj, here it is for your reference https://github.com/aws-controllers-k8s/code-generator/pull/190
The code it generates is the same but the way to enable it has changed. From setting `primary_identifier_field_name: ARN` to setting it as `is_arn_primary_key: true`

surajkota: Sg, I missed the diff b/w is_primary_key: true to is_arn_primary_key: true

surajkota (approved):
/approve
/lgtm

PR #124, opened 2021-10-19
Unit tests for HyperParameterJobTuning

ryansteakley:
Description of changes:

- `sdk.go` coverage 71.3 %
  - This coverage % would be higher however there are fields in [HPO](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sagemaker.html#SageMaker.Client.create_hyper_parameter_tuning_job), under `TrainingJobDefinition` that cannot be used only under `TrainingJobDefinitions`, trying to use some of the fields results in `validationExceptions `and tells the user to use `TrainingJobDefinitions`. Service requires user to supply values such as `HyperParameterRanges` and you cannot include `ParameterRanges` when only using `TrainingJobDefinition`. Appears to be a miss from the service team itself.
  - `hooks.go` coverage 100%
- Custom code coverage >95%


pkg/resource/hyper_parameter_tuning_job/manager_test_suite_test.go
```diff
@@ -100,15 +103,31 @@ func (d *testRunnerDelegate) Equal(a acktypes.AWSResource, b acktypes.AWSResourc
 	ac := a.(*resource)
 	bc := b.(*resource)
 	// Ignore LastTransitionTime since it gets updated each run.
-	opts := []cmp.Option{cmpopts.EquateEmpty(), cmpopts.IgnoreFields(ackv1alpha1.Condition{}, "LastTransitionTime")}
+	opts := []cmp.Option{cmpopts.EquateEmpty(),
+		cmpopts.IgnoreFields(ackv1alpha1.Condition{}, "LastTransitionTime"),
+		cmpopts.IgnoreFields(svcapitypes.HyperParameterTrainingJobSummary{}, "CreationTime"),
+		cmpopts.IgnoreFields(svcapitypes.HyperParameterTrainingJobSummary{}, "TrainingStartTime"),
+		cmpopts.IgnoreFields(svcapitypes.HyperParameterTrainingJobSummary{}, "TrainingEndTime")}
```
mbaijal, line 110: Values that are coming from the server side should be ok to compare, why do we need to ignore them ?

PR #134, opened 2021-11-09
Remove YamlEqual method to remove duplicate work and bash commands

ryansteakley:
Description of changes:
Remove yaml equality check from `util.go` as it is not necessary if we check the spec like we check the status. This will also remove the usage of bash commands such as `sed`, `diff`, and `cat`.


surajkota (approved):
/approve
/lgtm

PR #141, opened 2021-11-24
Test cross-region support

ryansteakley:
Description of changes:

-  Test ACKs cross region support so that it does not need to be tested manually every release

### DIFF
--- a/test/e2e/__init__.py
+++ b/test/e2e/__init__.py
@@ -35,19 +35,9 @@ bootstrap_directory = Path(__file__).parent
 resource_directory = Path(__file__).parent / "resources"
 
 
-def sagemaker_client():
-    return boto3.client("sagemaker")
-
-
-def cross_region_sagemaker_client():
-    return boto3.client("sagemaker", region_name=get_cross_region())
-
-
-def get_cross_region():
-    region = get_region()
-    if region != "us-west-2":
-        return "us-west-2"
-    return "us-east-1"
+def sagemaker_client(region: str = None):
+    region = region or get_region()
+    return boto3.client("sagemaker", region_name=region)
 
 
 def create_sagemaker_resource(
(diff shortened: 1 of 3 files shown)


test/e2e/resources/cross_region_model.yaml
```diff
@@ -0,0 +1,22 @@
...
+spec:
+  modelName: $MODEL_NAME
+  primaryContainer:
+    containerHostname: xgboost
+    image: $XGBOOST_V1_IMAGE_URI
+    environment:
+      my_var: my_value
+      my_var2: my_value2
+  executionRoleARN: $SAGEMAKER_EXECUTION_ROLE_ARN
+  tags:
```
surajkota, line 16: nit: remove tag, even env

test/e2e/tests/test_cross_region.py
```diff
@@ -0,0 +1,105 @@
...
+        return None
+
+
+@service_marker
+class TestCrossRegionModel:
+    def test_create_cross_region_model(self, cross_region_model):
+        (reference, resource) = cross_region_model
+        assert k8s.get_resource_exists(reference)
+
+        cross_region_sagemaker_client = boto3.client(
```
surajkota, line 88: make [this method](https://github.com/aws-controllers-k8s/sagemaker-controller/blob/main/test/e2e/__init__.py#L37) parameterized instead?

test/e2e/tests/test_cross_region.py
```diff
@@ -0,0 +1,105 @@
...
+            "sagemaker", region_name=get_cross_region()
+        )
+        model_name = resource["spec"].get("modelName", None)
+        model_desc = get_sagemaker_cross_region_model(
```
surajkota, line 92: same make this parameterized, take client as a param

client = param_client or sagemaker_client()

surajkota (changes requested):

test/e2e/tests/test_cross_region.py
```diff
@@ -0,0 +1,105 @@
...
+from e2e.common import config as cfg
+
+
+def get_cross_region():
+    region = get_region()
+    if region != "us-west-2":
+        region = "us-west-2"
+    else:
+        region = "us-east-1"
+    return region
```
A-Hilaly, line 41: Maybe it simpler to read if you just do
```python
def get_cross_region():
    region = get_region()
    if region == "us-west-2":
        return "us-west-2"
    return "us-east-1"
```

ryansteakley: yeah this looks cleaner, is a != if I'm correct however.

test/e2e/__init__.py
```diff
@@ -38,6 +39,17 @@ def sagemaker_client():
     return boto3.client("sagemaker")
 
 
+def cross_region_sagemaker_client():
```
surajkota, line 42: can instead modify the existing method?

```
def sagemaker_client(region: str = None):
    region = region or get_region()
    return boto3.client("sagemaker", region_name=region)
```


```
def get_sagemaker_model(config_name: str, sm_client = None):
  sm_client = sm_client or sagemaker_client()
  ...
```

so now you can create the client in the test file and pass it to these methods

test/e2e/__init__.py
```diff
@@ -38,6 +39,17 @@ def sagemaker_client():
...
+    return boto3.client("sagemaker", region_name=get_cross_region())
+
+
+def get_cross_region():
```
surajkota, line 46: +1
I think this method is better suited to be in the test file itself

surajkota (changes requested):
few nits

test/e2e/__init__.py
ryansteakley: Sounds good, If there is any reason for using cross_region in other tests in the future we can move it back here again

surajkota (approved):
/approve
/lgtm

PR #144, opened 2021-12-23
Increasing time out for SageMaker HPO tests

goswamig:
HPO jobs are timing out recently. Bumping the timeout for hpo jobs.

Error:
```
tests/test_hpo.py:121: AssertionError
------------------------------ Captured log call -------------------------------
INFO     root:resource.py:311 Condition ACK.ResourceSynced has status False, continuing...
ERROR    root:__init__.py:109 Wait for status: Completed timed out. Actual status: InProgress'
```

### DIFF
--- a/test/e2e/tests/test_hpo.py
+++ b/test/e2e/tests/test_hpo.py
@@ -91,7 +91,7 @@ class TestHPO:
         self,
         reference: k8s.CustomResourceReference,
         expected_status: str,
-        wait_periods: int = 45,
+        wait_periods: int = 90,
         period_length: int = 30,
     ):
         return wait_for_status(
@@ -106,7 +106,7 @@ class TestHPO:
         self,
         hpo_job_name,
         expected_status: str,
-        wait_periods: int = 45,
+        wait_periods: int = 90,
         period_length: int = 30,
     ):
         return wait_for_status(


surajkota:
@goswamig can you also decrease [this](https://github.com/aws-controllers-k8s/sagemaker-controller/blob/main/test/e2e/resources/xgboost_hpojob.yaml#L13) to 1?

Thx

goswamig:
> @goswamig can you also decrease [this](https://github.com/aws-controllers-k8s/sagemaker-controller/blob/main/test/e2e/resources/xgboost_hpojob.yaml#L13) to 1?
> 
> Thx

why is that needed ? HPO job with one training job does not make sense to me.

surajkota:
num training jobs is set to 2 and max parallel jobs is set to 1 which means service is running training job serially and the test might be timing out when service side waits for instances to be available(this is my suspect because the training job itself runs for 3-5 mins and the job is not timing out). Reducing the number of jobs might help with this

Since the objective of this test is to check the APIs of hyper_parameter_training_job, we are checking that passing an int in this parameter is working. rest is service side detail

goswamig:
HPO job with one training job does not make any difference wrt one training job. HPO runs multiple jobs and selects best parameters. The timeout should fix it, assuming its sufficient.

surajkota:
> HPO job with one training job does not make any difference wrt one training job

max num of jobs is bound by the resource limit specified by user

[boto3 ref](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sagemaker.html#SageMaker.Client.create_hyper_parameter_tuning_job
): _MaxNumberOfTrainingJobs (integer) -- [REQUIRED]
The maximum number of training jobs that a hyperparameter tuning job can launch_

Example below from one of the runs, there are 2 training jobs and the second one did not start till 20 mins after first job. 
Screenshot 2021-12-27 at 4 28 49 PM

first job was waiting around for instances to be ready
Screenshot 2021-12-27 at 4 33 48 PM

the situation becomes worse if one of the job fails because the service will retry it and test will timeout. (these are less frequent compared to training job taking long time to start)

goswamig:
If the job takes 3-5 mins why it was passing for 1 mins only before ? Did we look into it before ?
I disagree that we only check for api, if that’s the case you should not even have a test waiting for completion right?

Since I disagree on above point I still don’t see a point how a single job test HPO feature .

The intention of change is to allow enough time for training job to finish. If it’s an infra issue on service side, service  should have alarm and we should not do anything to cope with it.

goswamig:
Check the definition of HPO jobs [here](https://docs.aws.amazon.com/sagemaker/latest/dg/sagemaker-mkt-algo-tune.html)

A hyperparameter tuning job finds the best version of a model by **running many training jobs** on your dataset using the algorithm and ranges of hyperparameters that you specify.

surajkota:
> Check the definition of HPO jobs [here](https://docs.aws.amazon.com/sagemaker/latest/dg/sagemaker-mkt-algo-tune.html)
> 
> A hyperparameter tuning job finds the best version of a model by **running many training jobs** on your dataset using the algorithm and ranges of hyperparameters that you specify.

This is a service side detail and the service allows user to control the number of jobs to run given the parameter ranges. It will output the best parameters from the runs it was allowed to create. If the user sets resource limit to 1 it will run only one experiment and vice versa

> Since I disagree on above point I still don’t see a point how a single job test HPO feature.

Let me ask the question in a different way. Is there any reason you think the behaviour of controller is different when `maxNumberOfTrainingJobs` is `1` vs `n`?
**Note**: the controller is not creating or tracking the child training jobs created by the parent HPO job. This is done on service side. The controller only looks at [HyperParameterTuningJobStatus](https://github.com/aws-controllers-k8s/sagemaker-controller/blob/main/pkg/resource/hyper_parameter_tuning_job/hooks.go#L41) to decide if it needs to requeue.

> If the job takes 3-5 mins why it was passing for 1 mins only before ? Did we look into it before ?

In general, the job takes ~10mins given the instance startup time, dataset download etc. and so we started with 15 mins timeout. [Increased it to 30](https://github.com/aws-controllers-k8s/sagemaker-controller/pull/117) when we saw longer start up times on the training jobs(maybe 5-10% of the time). But we still find it is not sufficient because there are 2 training job and not one. Hence I think it is reasonable to scale down the instances instead of waiting for 45-60 mins for this test to complete. FYI: Wait time for single training job is set to [30](https://github.com/aws-controllers-k8s/sagemaker-controller/pull/143) because of similar reason.

> I disagree that we only check for api, if that’s the case you should not even have a test waiting for completion right?

I was focussing on the role of using this parameter under discussion rather than whole controller. By using this parameter in the test, we are verifying that we are able to take [this int64](https://github.com/aws-controllers-k8s/sagemaker-controller/blob/eee3470a4d9d6486e179417e086283c278a6bd11/apis/v1alpha1/types.go#L1998) from the CRD type and construct a [valid input for create in aws-go-sdk](https://github.com/aws-controllers-k8s/sagemaker-controller/blob/eee3470a4d9d6486e179417e086283c278a6bd11/pkg/resource/hyper_parameter_tuning_job/sdk.go#L1045). The actual number does not change the controller logic.

mbaijal:
I don't think either of these is the right fix here - 
1. We have already increased the timeout several times to no avail. Should it really be taking 90minutes for 2 instances to spin up and complete 2 short trainingjobs even if all these steps happen sequentially? 
2. While I agree that testing 1 vs 2 trainingJobs for HPO does not impact the quality of our integ test in any way, IMO we should not be reducing the number of trainingJobs to fix this test since that feels more like a workaround than fix. 

For now, as discussed with @surajkota offline, it might be better to increase `parallelTrainingJobs` before we reduce the `maxNumberOfTrainingJobs` to ensure reduced runtime. 

Eventually, we should try and root cause why this is taking so long - 
  - Do we know if each of these runs on the same instance or different one - trying to understand why the bootstrapping time is so high.  
  - Can we make sure we have enough account limits in place 
  - Can we sync up with the SageMaker team about the expected runtimes ?

mbaijal (changes requested):
timeout has been increased multiple times before. Can you confirm with more details why an HPOjob is taking ~90 minutes.

ryansteakley:
closed in favor of https://github.com/aws-controllers-k8s/sagemaker-controller/pull/146

PR #150, opened 2022-01-24
Fix canary tests v0.3.0

akartsky:
1] Cherry-pick some of the fixes from main
2] For all notebook tests retry failed tests once

### DIFF
--- a/test/e2e/tests/test_notebook_instance.py
+++ b/test/e2e/tests/test_notebook_instance.py
@@ -27,7 +27,9 @@ from e2e import (
     sagemaker_client,
 )
 from e2e.replacement_values import REPLACEMENT_VALUES
+from flaky import flaky
 import random
+
 
 DELETE_WAIT_PERIOD = 16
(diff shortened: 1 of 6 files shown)


test/e2e/tests/test_notebook_instance.py
```diff
@@ -84,6 +86,7 @@ def get_notebook_instance_resource_status(reference: k8s.CustomResourceReference
     return resource["status"]["notebookInstanceStatus"]
 
 
+@flaky
```
surajkota, line 89: Great! how do you set num of retires? is there a default?

surajkota (approved):
approving assuming the default num retries < 2

test/e2e/tests/test_notebook_instance.py
akartsky: we can set it like this `@flaky(max_runs=3, min_passes=2)`

default is `max_runs=2, min_passes=1`

ryansteakley (approved):

PR #152, opened 2022-02-08
Upgrade to ACK runtime v0.16.5 and aws-sdk-go v1.42.0

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1167

Description of changes:
* Updates ACK runtime from v0.15.2 to v0.16.5 . This upgrades results in aws-sdk-go upgrade to v1.42.0 and controller-runtime update to v0.11.0 as well
* Ignored the empty struct shape _RSessionAppSettings_ and new resource _StudioLifecycleConfig_ in generator.yaml file
* Updated the mocks to generated from new `aws-sdk-go` version. Also fixed some underlying bug in creating mocks from Makefile


generator.yaml
```diff
@@ -865,3 +866,5 @@ ignore:
       - Workteam
     operations:
       - UpdateTrainingJob
+    shape_names:
+      - RSessionAppSettings
```
jaypipes, line 870: I always find it useful to add a comment above these types of additions, just to let readers (or future me...) know why I did certain things. For example, here, I'd put something to the effect of:

```yaml
      # This is an empty struct that causes generation errors
```

jaypipes (approved):
:+1:

surajkota:
@vijtrip2 thanks this PR. I am putting this on hold since its updating a bunch of CRDs as a result of updating go sdk.

latest go sdk can bring in features which might need changes to the controller logic. e.g. I see a retry option in trainingJob CRD and new deployment options and rollback settings in endpoint CRD. Although, the changes required for those features will be separate PRs, we want have to a closer look and atleast estimate the effort needed for those.

Is it possible to stay on the previous go sdk version?

vijtrip2:
> Is it possible to stay on the previous go sdk version?

Not recommended. You can specify aws-sdk-go version when performing code-generation but the aws-sdk-go version in go.mod file will be minimal v1.42.0 because of ACK runtime.

By specifying custom aws-sdk-go in code-generation, the aws-sdk-go version in go.mod is v1.42.0 but the generated types will be from aws-sdk-go version from code-generation option. This will cause inconsistencies in your mocks etc and maybe some more hidden issues. But it's possible.

surajkota:
is this a new requirement that the controller and runtime go sdk version need to match?
sagemaker controller was always using a different go sdk version from the runtime.

Can you provide an example where it would lead to inconsistencies? because the existing fields do not change, there might be new shapes for which code gen might have additional logic

as far as runtime is concerned, unless the go sdk session object changes(it would be caught in e2e tests) it should be safe?

vijtrip2:
> is this a new requirement that the controller and runtime go sdk version need to match? sagemaker controller was always using a different go sdk version from the runtime.

It a golang feature where you can't have package version lesser than package version in your dependency. You can have aws-sdk-go version greater than what's present in ACK runtime go.mod


> Can you provide an example where it would lead to inconsistencies? because the existing fields do not change, there might be new shapes for which code gen might have additional logic

One example I can think of is if the new aws-sdk-go version(greater) adds a required field and you generate the controller using old aws-sdk-go version(lesser). The generated code will compile but fail at execution time because aws-sdk-go version used at execution time will be greater version

vijtrip2:
@ryansteakley is going to raise a separate PR with new resources ignored.

/close

PR #153, opened 2022-02-25
Upgrade to ACK runtime v0.16.5 and aws-sdk-go v1.42.0 

ryansteakley:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1167

Description of changes:

- Updates ACK runtime from v0.15.2 to v0.16.5 . This upgrades results in aws-sdk-go upgrade to v1.42.0 and controller-runtime update to v0.11.0 as well
-  Ignored the empty struct shape `RSessionAppSettings `and new resource `StudioLifecycleConfig `in generator.yaml file
-  Ignored the field `RetryStrategy `for the resources TrainingJob and HyperparameterTuning job in generator.yaml file; suspect that it may need custom-code/implementation will need further testing.
- Ignored the field `DeploymentConfig `for the resource Endpoint in generator.yaml file; suspect that it may need custom-code/implementation will need further testing.
-  Updated the mocks to generated from new aws-sdk-go version. Also fixed some underlying bug in creating mocks from Makefile
-  Removed check for ack-terminal is set to False in `test/e2e/tests/test_endpoint.py` since the field will now be set to None on requeues instead of back to False
- Call `customSetOutput` in `sdkUpdate` for ModelPackage to set the synced conditions `pkg/resource/model_package/sdk.go`
- Added unit-tests for new late-initialized field in Notebook


apis/v1alpha1/generator.yaml
```diff
@@ -505,6 +505,8 @@ resources:
         template_path: model_package/set_resource_name_as_arn.go.tpl
       sdk_update_pre_build_request:
         template_path: common/sdk_update_pre_build_request.go.tpl
+      sdk_update_post_set_output:
+        code: rm.customSetOutput(&resource{ko})
```
RedbackThomson, line 509: In the latest version of the code-generator we can do this with `synced.when.in[]` in the generator config. Check out how we do it in the [EKS controller](https://github.com/aws-controllers-k8s/eks-controller/blob/main/generator.yaml#L106-L110)

pkg/resource/model_package/sdk.go
```diff
@@ -937,13 +976,20 @@ func (rm *resourceManager) newUpdateRequestPayload(
...
+		}
+		res.SetCustomerMetadataProperties(f1)
+	}
 	if r.ko.Spec.ModelApprovalStatus != nil {
 		res.SetModelApprovalStatus(*r.ko.Spec.ModelApprovalStatus)
 	}
 	if r.ko.Status.ACKResourceMetadata != nil && r.ko.Status.ACKResourceMetadata.ARN != nil {
 		res.SetModelPackageArn(string(*r.ko.Status.ACKResourceMetadata.ARN))
-	} else {
-		res.SetModelPackageArn(rm.ARNFromName(*r.ko.Spec.ModelPackageName))
```
RedbackThomson, line 946: Did we want to lose this?

RedbackThomson (changes requested):
Great job updating all of these tests and getting things back into a working order. This would probably be 10x easier if we iterate more quickly on sdk versions, rather than having a multi-thousand line pull request.

I have left a comment regarding the new hook you are using for patching the `ResourceSynced` status. This can now be done using generated code with the latest runtime. It might be worth it for you to bump the runtime again, up to the newest version, and regenerate with that. What is a few hundred more lines anyway?

pkg/resource/model_package/sdk.go
ryansteakley: I believe this came as part of the automated codegeneration. Since this is in `sdkUpdate` and only a versioned ModelPackage can be updated I believe it is okay to lose this. Reason being that a `VersionedModelPackage` doesn't have a `ModelPackageName` and we set the `ModelPackageName` as the `ARN` in `sdkCreate` so this else statement is redundant.

RedbackThomson: I'm happy if you're happy

RedbackThomson:
Pushing this through so SageMaker can be updated to support the latest runtime
/lgtm

PR #158, opened 2022-06-17
Update ACK runtime to v0.19.1

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1351

Description of changes:
* Update ACK runtime to v0.19.1
* ignore ack system tags during tag assertion. Unable to use acktest from test-infra because of key name difference. Key vs key
* ignore tags for ModelPackage because tags are only supported for ModelPackageGroup.


A-Hilaly (approved):
/lgtm

generator.yaml
```diff
@@ -540,6 +540,8 @@ resources:
       Tags:
         compare:
           is_ignored: true
+    tags:
```
surajkota, line 543: @vijtrip2 could you describe what this means? since we already set ignore for comparing Tags. is this a breaking change?

test/e2e/tags.py
```diff
@@ -0,0 +1,25 @@
...
+
+from typing import List
+
+ACK_SYSTEM_TAG_PREFIX = "services.k8s.aws/"
+
+def clean(tags: List[object]) -> List[object]:
+    """Returns supplied tags collection, stripped of ACK system tags.
+    """
+    return [
+        t for t in tags if not t['key'].startswith(ACK_SYSTEM_TAG_PREFIX)
```
surajkota, line 24: Instead of ignoring the ACK_SYSTEM_TAGS, we should assert the tag Keys are present. Atleast the keys since value can keep changing

surajkota:
@vijtrip2 Thanks for the PR and updating sagemaker-controller. Few questions for clarification, Thanks

generator.yaml
vijtrip2: No, it is not a breaking change. This configuration will just not add ACK system tags for this resource.
When I was fixing the e2e tests, i saw the following error message, and ignored setting ACK system tags for this resource.

```
Tags are not supported in Model Package versions. Please add them to the Model Package Group.                                                                                                    │
│                            status code: 400, request id: ee23b541-c579-4f9e-9bc1-8ef1b081112e
```

test/e2e/tags.py
vijtrip2: yeah, I wish to do it for all the controllers together from a common test-infra lib.
Currently, I was just fixing the tag assertion failures.

PR #175, opened 2022-10-20
Feature: Support RetryStrategy/InstanceGroups

ananth102:
Note: Some modifications are made to Hyper parameter tuning, this is normal as Hyper parameter tuning consists of training jobs.

Description of changes:
generator.yaml - commented out shapes. Added late initialize to ResourceConfig.InstanceCount and it's HPO equivalent. 


generator.yaml
```diff
@@ -188,6 +188,9 @@ resources:
       OutputDataConfig.KMSKeyID:
         late_initialize:
           min_backoff_seconds: 5
+      ResourceConfig.InstanceCount:
```
surajkota, line 191: is this change in behaviour?

what happens if user is using instance groups only?

surajkota (commented):
shall we add the retryStrategy to existing e2e tests? free of cost fault tolerance

generator.yaml
ananth102: Added retry strategy to integration tests and added unit tests.

>is this change in behavior?
Yes, instance count is optional because the user can specify Instance group. However the server side default of instanceCount is 0.
> what happens if user is using instance groups only?
Instance count will be late initialized to 0.

surajkota (changes requested):
missing handling `TrainingJobDefinitions`.ResourceConfig.InstanceCount in HPO

also reduce the retry count to 3

generator.yaml
surajkota: this is a very strange behaviour

ananth102:
> missing handling TrainingJobDefinitions.ResourceConfig.InstanceCount in HPO

ACK, I've been trying to test that but I keep getting credential errors on sagemaker.

ananth102:
/hold
Merge this after #177. Also still need to debug TrainingJobDefinitions.ResourceConfig.InstanceCount.

ananth102:
/unhold
TrainingJobDefinitions.ResourceConfig.InstanceCount works

surajkota:
/ok-to-test

looks ok, do you want to merge 177 first?

ananth102:
\hold

ryansteakley (approved):

PR #179, opened 2022-11-18
WIP: Feature: UpdateTrainingJob

ananth102:
Description of changes:

Update trainingjob related, Unit tests will be included in a separate PR since I anticipate them to be large. The unit tests changes in this PR are only for compatibility.

Update trainingjob related custom code changes:
`generator.yaml` - Included Update Training job and its hooks
`pkg/resource/training_job/custom_delta.go` - Added bounds check for server side default and also a post delta function that removes Profiler related things from the delta if the user decides to disable the profiler.
`pkg/resource/training_job/custom_set_update_input.go` - Modifies input for update, needed because the API functions like a POST instead of a PUT.
`pkg/resource/training_job/custom_update_conditions.go` - Sets the resource to terminal if [ACK_SM] is included in the error.
`pkg/resource/training_job/hooks.go` - Various custom functions to either check if an update cannot happen or can happen in the future.
`pkg/resource/training_job/testdata/*` - Changes for compatibility.


pkg/resource/training_job/custom_set_update_input.go
```diff
@@ -0,0 +1,125 @@
...
+// Recreates input and sets disable profiler to true
+func handleProfilerRemoval(input *svcsdk.UpdateTrainingJobInput) {
+	input.SetProfilerRuleConfigurations(nil)
+	profilerConfig := svcsdk.ProfilerConfigForUpdate{}
+	profilerConfig.SetDisableProfiler(true)
+	input.SetProfilerConfig(&profilerConfig)
+}
+
+// Sagemaker and kubernetes types are not the same so the input has to be reconstructed.
+func convertProfileRuleType(rule *smv1alpha.ProfilerRuleConfiguration) *svcsdk.ProfilerRuleConfiguration {
```
ananth102, line 95: Is there a way to implicitly/explicity convert the Kubernetes type to AWS API type. Dont want to have this function as it is not maintainable.

pkg/resource/training_job/sdk.go
```diff
@@ -1060,6 +1060,9 @@ func (rm *resourceManager) sdkUpdate(
 	if warmpool_diff && profiler_diff {
 		return latest, errors.New("[ACK_SM] Cannot update Warm pool and Profiler at the same time.")
 	}
+	if !warmpool_diff && !profiler_diff {
```
ananth102, line 1063: Added this to prevent an infinite loop, but this might be a problem if there is a server side default update/bug. Is there a better way to handle this?

pkg/resource/training_job/sdk.go
```diff
@@ -1060,6 +1060,9 @@ func (rm *resourceManager) sdkUpdate(
...
+		return latest, errors.New("[ACK_SM] Only Warm Pool or Profiler can be updated")
```
A-Hilaly, line 1064: >Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context. That is, use fmt.Errorf("something bad") not fmt.Errorf("Something bad"), so that log.Printf("Reading %s: %v", filename, err) formats without a spurious capital letter mid-message. This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.

from https://github.com/golang/go/wiki/CodeReviewComments#error-strings

A-Hilaly: Also the `[ACK_SM]` is just adding noise to the logs. Users already see this information in the controller logs, along with reconciller name, namespace and a lot of other stuff.

ananth102: Good point, I will use lowercase errors.

> Also the [ACK_SM] is just adding noise to the logs.

[ACK_SM] is used as a terminal prefix in UpdateConditions(https://github.com/aws-controllers-k8s/sagemaker-controller/pull/179/files#diff-06bfc8acec53697c68f4ed8d7e1f0192ec628506edeb9a769036611a78dc2ae1). The alternative I considered manually iterating through all possible terminal errors(which isn't maintainable, might cause errors if the sagemaker server errors change). I will rename that prefix to something like acksm_terminal: as its a bit more clear.

RedbackThomson: Rather than using matching for this, we already have a `TerminalError` type, which the common runtime detects and applies as a terminal condition - https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/errors/error.go#L104-L106

pkg/resource/training_job/sdk.go
ananth102: Using DifferenceExcept does not work well with unit tests as unit tests do a delta after the update as opposed to before.

pkg/resource/training_job/sdk.go
```diff
@@ -1087,9 +1113,149 @@ func (rm *resourceManager) sdkUpdate(
...
+		if err := customSetOutputUpdateWarmpool(latest); err != nil {
+			return nil, err
+		}
+	}
+	if profiler_diff {
+		if up_err := customSetOutputUpdateProfiler(latest); up_err != nil {
+			return nil, up_err
+		}
+		input.SetResourceConfig(nil)
+		if profilerRemovalCheck(desired, latest) {
```
surajkota, line 1151: rename to isProfilerRemoved

pkg/resource/training_job/sdk.go
```diff
@@ -1087,9 +1113,149 @@ func (rm *resourceManager) sdkUpdate(
...
+	if ko.Status.ACKResourceMetadata == nil {
+		ko.Status.ACKResourceMetadata = &ackv1alpha1.ResourceMetadata{}
+	}
+	if resp.TrainingJobArn != nil {
+		arn := ackv1alpha1.AWSResourceName(*resp.TrainingJobArn)
+		ko.Status.ACKResourceMetadata.ARN = &arn
+	}
+
+	rm.setStatusDefaults(ko)
+	customSetOutputPostUpdate(ko, delta)
```
surajkota, line 1181: call a readOne instead of this custom checks and set status from it. See the sdkdelete method for reference

pkg/resource/training_job/sdk.go
```diff
@@ -1087,9 +1113,149 @@ func (rm *resourceManager) sdkUpdate(
...
+	}
+	if profiler_diff {
+		if up_err := customSetOutputUpdateProfiler(latest); up_err != nil {
+			return nil, up_err
+		}
+		input.SetResourceConfig(nil)
+		if profilerRemovalCheck(desired, latest) {
+			handleProfilerRemoval(input)
+		} else {
+			inp_err := customSetUpdateInput(desired, latest, delta, input)
```
surajkota, line 1154: can we move these methods handleProfilerRemoval, customSetUpdateInput etc. to hooks.go?

again, please make these methods part of resource manage class

pkg/resource/training_job/custom_set_update_input.go
surajkota: Unfortunately I dont think so

pkg/resource/training_job/custom_set_update_input.go
```diff
@@ -0,0 +1,130 @@
...
+	profilerConfig := svcsdk.ProfilerConfigForUpdate{}
+	profilerConfig.SetDisableProfiler(true)
+	input.SetProfilerConfig(&profilerConfig)
+}
+
+// convertProfileRuleType converts the kubernetes object ProfilerRuleConfiguration into
+// a type that is compatible with the AWS API.
+// Sagemaker and kubernetes types are not the same so the input has to be reconstructed.
+func convertProfileRuleType(rule *smv1alpha.ProfilerRuleConfiguration) *svcsdk.ProfilerRuleConfiguration {
+	smRule := &svcsdk.ProfilerRuleConfiguration{}
```
surajkota, line 101: rename to rule, all of the code is related to SM

pkg/resource/training_job/custom_set_update_input.go
```diff
@@ -0,0 +1,130 @@
...
+// of training job which is not evident from API response
+
+package training_job
+
+import (
+	"errors"
+
+	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
+	ackerr "github.com/aws-controllers-k8s/runtime/pkg/errors"
+	smv1alpha "github.com/aws-controllers-k8s/sagemaker-controller/apis/v1alpha1"
```
surajkota, line 24: please follow other files to maintain consistency

pkg/resource/training_job/custom_set_update_input.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+// customSetUpdateInput modifies the input of UpdateTrainingJob.
+// Three conditions:
+// 1. Customer updates both profiler parameters: Recreate the input for profiler Rule.
+// 2. Customer only updates Profiler Config: Set the profiler rule configuration to nil to avoid validation error.
+// 3. Customer only updates Rule Configurations: Recreate the input for profiler Rule and set Profiler config to nil.
+//	  safer to do this because the "only add" behavior might reappear.
+func customSetUpdateInput(desired *resource, latest *resource, delta *ackcompare.Delta, input *svcsdk.UpdateTrainingJobInput) error {
+	if delta.DifferentAt("Spec.ProfilerConfig") && delta.DifferentAt("Spec.ProfilerRuleConfigurations") {
```
surajkota, line 35: checks have duplication? does the following work?

```
if !delta.DifferentAt("Spec.ProfilerConfig") {
  input.SetProfilerConfig(nil)
}

if !delta.DifferentAt("Spec.ProfilerRuleConfigurations") {
  input.SetProfilerRuleConfigurations(nil)
} else {
  err := handleProfilerRuleConfig(desired, latest, input)
}
```

pkg/resource/training_job/custom_set_update_input.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	return nil
+}
+
+// handleProfilerRuleConfig sets the input of the ProfilerRuleConfiguration so that
+// it is compatible with the sagemaker API.
+// Update training job is post operation wrt to the profiler parameters.
+// Because of this only NEW rules can be specified.
+// In this function we check to see if any new profiler configurstions have been added.
+func handleProfilerRuleConfig(desired *resource, latest *resource, input *svcsdk.UpdateTrainingJobInput) error {
```
surajkota, line 56: `buildProfilerRuleConfigUpdateInput`

pkg/resource/training_job/custom_set_update_input.go
```diff
@@ -0,0 +1,130 @@
...
+	profilerRuleDesired := desired.ko.Spec.ProfilerRuleConfigurations
+	profilerRuleLatest := latest.ko.Spec.ProfilerRuleConfigurations
+
+	if ackcompare.IsNil(profilerRuleDesired) {
```
surajkota, line 60: this is already handled when you check if profiler needs to be disabled?

pkg/resource/training_job/custom_set_update_input.go
```diff
@@ -0,0 +1,130 @@
...
+		return nil
+	}
+	if len(profilerRuleDesired) < len(profilerRuleLatest) {
+		return ackerr.NewTerminalError(errors.New("cannot remove a profiler rule."))
+	}
+
+	ruleMap := map[string]int{}
+	profilerRuleInput := []*svcsdk.ProfilerRuleConfiguration{}
+	for _, rule := range profilerRuleLatest {
+		if ackcompare.IsNotNil(rule) && ackcompare.IsNotNil(rule.RuleConfigurationName) {
```
surajkota, line 73: why would the rule or the rule name be nil? rule name is required field

surajkota (changes requested):
please simplify the logic, added comments

pkg/resource/training_job/hooks.go
```diff
@@ -57,11 +79,120 @@ func (rm *resourceManager) customSetOutput(r *resource) {
...
+// customSetOutputUpdateProfiler decides whether the training job is ready/eligible for update
+// depending on the status.
+func customSetOutputUpdateProfiler(r *resource) error {
+	trainingSecondaryStatus := r.ko.Status.SecondaryStatus
+	trainingJobStatus := r.ko.Status.TrainingJobStatus
+	if ackcompare.IsNotNil(trainingSecondaryStatus) && *trainingSecondaryStatus == svcsdk.SecondaryStatusStarting {
+		return requeueBeforeUpdateStarting
+	}
+	if ackcompare.IsNotNil(trainingJobStatus) {
+		for _, terminalStatus := range TrainingJobTerminalProfiler {
```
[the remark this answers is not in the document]
ananth102: wont the controller constantly reconcile in that case?

config/crd/bases/sagemaker.services.k8s.aws_hyperparametertuningjobs.yaml
```diff
@@ -431,6 +431,9 @@ spec:
                         type: integer
                       instanceType:
                         type: string
+                      keepAlivePeriodInSeconds:
```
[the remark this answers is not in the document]
ananth102: the code generator cannot do that unfortunately since hpo/trainingjob share the resourceconfig in types.go

pkg/resource/training_job/hooks.go
```diff
@@ -57,11 +79,120 @@ func (rm *resourceManager) customSetOutput(r *resource) {
...
+	if ackcompare.IsNotNil(trainingJobStatus) {
+		if *trainingJobStatus == svcsdk.TrainingJobStatusInProgress {
+			return false
+		}
+		if *trainingJobStatus == svcsdk.TrainingJobStatusCompleted {
+			if ackcompare.IsNotNil(latest.ko.Status.WarmPoolStatus) {
+				wp_modifying := svccommon.IsModifyingStatus(latest.ko.Status.WarmPoolStatus.Status, &WarmPoolModifyingStatuses)
+				return !wp_modifying
+			} else {
+				return false // Sometimes the API (briefly) does not return the WP status even if it completes.
```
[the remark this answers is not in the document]
surajkota: Clarified, there is a delay between when the trainingjob completes and service starts returning the warm pool status. Once the status is available it is up to date

pkg/resource/hyper_parameter_tuning_job/custom_delta.go
```diff
@@ -31,6 +31,30 @@ func customSetDefaults(
...
 		}
+		// TODO: Remove the block below.
+		// The server side default of KeepAlivePeriodInSeconds is nil, when launching a HPO job.
+		// The code generator currently cannot ignore the field path for resourceConfig.KeepAlivePeriodInSeconds
+		// without also ignoring Trainingjob. This block below should be removed once the code generator supports
+		// removing fields like resourceConfig.KeepAlivePeriodInSeconds
+		if ackcompare.IsNotNil(a.ko.Spec.TrainingJobDefinition) && ackcompare.IsNotNil(b.ko.Spec.TrainingJobDefinition) {
+			if ackcompare.IsNotNil(a.ko.Spec.TrainingJobDefinition.ResourceConfig) && ackcompare.IsNotNil(b.ko.Spec.TrainingJobDefinition.ResourceConfig) {
+				if ackcompare.IsNotNil(a.ko.Spec.TrainingJobDefinition.ResourceConfig.KeepAlivePeriodInSeconds) && ackcompare.IsNil(b.ko.Spec.TrainingJobDefinition.ResourceConfig.KeepAlivePeriodInSeconds) {
+					a.ko.Spec.TrainingJobDefinition.ResourceConfig.KeepAlivePeriodInSeconds = nil
```
surajkota, line 42: can we ignore this shape from HPO since user cannot specify it?

ananth102: no, because the shape is shared with training job.

pkg/resource/training_job/sdk.go
ananth102: move both of those methods to hooks.go, the other methods in customsetinput.go modify the input so its better for them to stay there.

pkg/resource/training_job/custom_set_update_input.go
ananth102: Yes but smRule is a sagemaker api type and rule is a kubernetes object type.

templates/training_job/sdk_update_post_build_request.go.tpl
```diff
@@ -0,0 +1,36 @@
...
+	}
+}
+if profiler_diff {
+	if up_err := rm.customSetOutputUpdateProfiler(latest); up_err != nil {
+		return nil, up_err
+	}
+	input.SetResourceConfig(nil)
+	if rm.isProfilerRemoved(desired, latest) {
+		rm.handleProfilerRemoval(input)
+	} else{
```
ryansteakley, line 30: indentation a little weird on else statement?

pkg/resource/training_job/custom_set_update_input.go
```diff
@@ -0,0 +1,104 @@
...
+// Because of this only NEW rules can be specified.
+// In this function we check to see if any new profiler configurstions have been added.
+func buildProfilerRuleConfigUpdateInput(desired *resource, latest *resource, input *svcsdk.UpdateTrainingJobInput) error {
+	profilerRuleDesired := desired.ko.Spec.ProfilerRuleConfigurations
+	profilerRuleLatest := latest.ko.Spec.ProfilerRuleConfigurations
+
+	if ackcompare.IsNil(profilerRuleLatest) {
+		return nil
+	}
+	if len(profilerRuleDesired) < len(profilerRuleLatest) {
```
ryansteakley, line 40: Dumb question, but if its possible that len(profilerRuleDesired) might be less than latest, can profilerRuleDesired be nil and this method still be called somehow? I'm assuming no

ananth102: No, if profilerRuleDesired is nil the profiler is removed. https://github.com/aws-controllers-k8s/sagemaker-controller/blob/d7f828a94c30f11f43dac1ce3ab472f657c68d6c/pkg/resource/training_job/sdk.go#L1165

pkg/resource/training_job/custom_delta.go
```diff
@@ -29,6 +29,10 @@ func customSetDefaults(
 
 	if ackcompare.IsNotNil(a.ko.Spec.ProfilerRuleConfigurations) && ackcompare.IsNotNil(b.ko.Spec.ProfilerRuleConfigurations) {
 		for index := range a.ko.Spec.ProfilerRuleConfigurations {
+			// Prevent out of bounds panics.
+			if index == len(a.ko.Spec.ProfilerRuleConfigurations) || index == len(b.ko.Spec.ProfilerRuleConfigurations) {
```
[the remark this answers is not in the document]
ananth102: Gone ahead an refactored it.

pkg/resource/training_job/hooks.go
```diff
@@ -57,11 +79,120 @@ func (rm *resourceManager) customSetOutput(r *resource) {
...
+	}
+
+}
+
+// customSetOutputUpdateWarmpool makes the controller requeue if there is an update and
+// the training job is still in InProgress
+func customSetOutputUpdateWarmpool(r *resource) error {
+	trainingJobStatus := r.ko.Status.TrainingJobStatus
+	if ackcompare.IsNotNil(trainingJobStatus) && *trainingJobStatus == svcsdk.TrainingJobStatusInProgress {
+		return requeueBeforeUpdate
```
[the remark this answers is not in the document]
ananth102: Using it somewhere else

pkg/resource/training_job/hooks.go
```diff
@@ -57,11 +79,120 @@ func (rm *resourceManager) customSetOutput(r *resource) {
...
+	// ACK OIDC is misconfigured (Terminal)
+	return true
+}
+
+// customSetOutputUpdateProfiler decides whether the training job is ready/eligible for update
+// depending on the status.
+func customSetOutputUpdateProfiler(r *resource) error {
+	trainingSecondaryStatus := r.ko.Status.SecondaryStatus
+	trainingJobStatus := r.ko.Status.TrainingJobStatus
+	if ackcompare.IsNotNil(trainingSecondaryStatus) && *trainingSecondaryStatus == svcsdk.SecondaryStatusStarting {
```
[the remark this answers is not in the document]
ananth102: done

pkg/resource/training_job/custom_set_update_input.go
ananth102: took it out

pkg/resource/training_job/custom_set_update_input.go
ananth102: works

pkg/resource/training_job/custom_set_update_input.go
surajkota: this is pretty basic check and is not robust, if a customer changes the rules instead of adding new ones or removed few of them, current logic will fail after the update, it will add the new rules to the job but still find a diff afterwards.

@ananth102 you already mentioned it in the comments but maybe worth clarifying the expectations further for future reference

I intentionally ignored it for now to keep the focus on adding rules when the customer wants to instead of covering all possibilities based on service behavior. We can do one simple thing to make it a bit more intelligent instead of comparing the length, mark the rules with 0 when you find from latest and mark them 1 when you find it in desired. After the loop, if all the ruleMap entries are not 1, error saying existing rules cannot be removed. wdyt?

also change the file name to custom_update if you plan on keeping it

ananth102: Sounds like there are four cases:
1. Customer adds rule (Happy)
2. Customer removes rule (terminal)
3. Customer removes some rules and adds other -> Controller would try/fail updates and eventually be set to terminal
4. Customer changes rule -> Controller would try/fail updates and eventually be set to terminal

That approach in your comment solves the first three cases. For the fourth one I can change the < in line 40 to <= and return a regular error instead of a terminal error. It will eventually terminate once the training job completes.

pkg/resource/training_job/custom_delta.go
```diff
@@ -60,3 +65,53 @@ func customSetDefaults(
...
+	return profilerConfigPresent && profilerRuleConfigPresent
+}
+
+// removeDelta Removes fieldName from the delta slice.
+// TODO: Replace when ack runtime can do this.
+func removeDelta(delta *ackcompare.Delta, fieldName string) {
+	differences := delta.Differences
+	for index, diff := range differences {
+		if diff.Path.Contains(fieldName) {
+			differences = append(differences[:index], differences[index+1:]...)
```
surajkota, line 112: index + 1 can be out of bounds?

ananth102: If I have s [a] -> slice | cap = 1 length = 1
s[1] isnt valid because the length of the slice is 1. But s[1:] is valid because the capacity is one.
As per [go spec](https://go.dev/ref/spec#Slice_expressions) "For slices, the upper index bound is the slice capacity cap(a) rather than the length."
s[2:] would be invalid because 2 > cap(s)
Differences is a slice. cap(differences) = differences[index+1:]

ananth102: >is it better if we ignore the default generated delta comparison and use reflect.DeepEqual when needed to compare both?

delta.go already uses deep [equal](https://github.com/aws-controllers-k8s/sagemaker-controller/blob/1989765a2bccadb1200da8b8411e7eece000991f/pkg/resource/training_job/delta.go#L232), the objective of this code is to remove the elements in the delta if the profiler is disabled. The alternative would be manually overwriting the desired object in pre compare which would require more boilerplate and could potentially have edge cases.

pkg/resource/training_job/custom_update.go
```diff
@@ -0,0 +1,138 @@
...
+	return nil
+}
+
+// markNonUpdatableRules returns a map containing the rules that are not eligible for update.
+// In addition it returns an error if a rule gets removed.
+func (rm *resourceManager) markNonUpdatableRules(profilerRuleDesired []*svcapitypes.ProfilerRuleConfiguration, profilerRuleLatest []*svcapitypes.ProfilerRuleConfiguration) (map[string]int, error) {
+	commonRulesMap := map[string]int{}
+	latestRulesMap := map[string]int{}
+	for _, rule := range profilerRuleLatest {
+		if ackcompare.IsNotNil(rule.RuleConfigurationName) {
```
surajkota, line 76: ruleConfiguration name is required, we cam drop this check?

pkg/resource/training_job/custom_delta.go
```diff
@@ -60,3 +65,53 @@ func customSetDefaults(
...
+		if *profilerStatus == "Disabled" && !userInitiatesProfilerCheck(desired) {
+			profilerDisabled = true
+		} else {
+			return
+		}
+	} else {
+		return
+	}
+	// TODO: Replace remove delta with an ack version when its natively supported
+	if profilerConfigDiff && profilerDisabled {
```
surajkota, line 91: move both of these inside the if block and drop the else in line 84 and 87

pkg/resource/training_job/custom_delta.go
```diff
@@ -60,3 +65,53 @@ func customSetDefaults(
...
+		removeDelta(delta, "Spec.ProfilerConfig")
+	}
+	if profilerRuleDiff && profilerDisabled {
+		removeDelta(delta, "Spec.ProfilerRuleConfigurations")
+	}
+}
+
+// userInitiatesProfilerCheck checks if the user enabled/re enabled the profiler.
+func userInitiatesProfilerCheck(desired *resource) bool {
+	profilerConfigPresent := ackcompare.IsNotNil(desired.ko.Spec.ProfilerConfig)
```
surajkota, line 101: can we just write the condition directly instead of creating vars, they are not reused

surajkota (changes requested):

templates/training_job/sdk_update_post_set_output.go.tpl
```diff
@@ -1,3 +1,8 @@
+observed, err := rm.sdkFind(ctx, latest)
+if err != nil {
+    return observed, err
+}
+desired.SetStatus(observed)
```
ryansteakley, line 5: this to get updated status on 1st update loop?

pkg/resource/training_job/custom_update.go
```diff
@@ -56,10 +59,41 @@ func (rm *resourceManager) buildProfilerRuleConfigUpdateInput(desired *resource,
...
+			latestRulesMap[*rule.RuleConfigurationName] = 0
+		}
+	}
+	for _, rule := range profilerRuleDesired {
+		if ackcompare.IsNotNil(rule.RuleConfigurationName) {
+			commonRulesMap[*rule.RuleConfigurationName] = 1
+		}
+	}
+	for _, val := range commonRulesMap {
+		// This means that there exists a rule in latest that is not present in desired
```
ryansteakley, line 87: This comment boils down to they tried/attempted to remove a profiler rule?

pkg/resource/training_job/hooks.go
```diff
@@ -57,11 +74,94 @@ func (rm *resourceManager) customSetOutput(r *resource) {
...
+	}
+	return nil
+}
+
+// isProfilerRemoved checks if the profiler was removed.
+func (rm *resourceManager) isProfilerRemoved(desired *resource, latest *resource) bool {
+	if ackcompare.IsNil(desired.ko.Spec.ProfilerRuleConfigurations) && ackcompare.IsNotNil(latest.ko.Spec.ProfilerRuleConfigurations) {
+		return true
+	}
+	if ackcompare.IsNil(desired.ko.Spec.ProfilerConfig) && ackcompare.IsNotNil(latest.ko.Spec.ProfilerConfig) {
```
ryansteakley, line 143: I see that in customsetUpdateInput one of the conditions is 1. Customer updates both profiler parameters....
So both ProfilerConfig and ProfilerRuleConfigurations can be updated. However in this code if one of these updated it will return. So in this case the logic for ProfilerConfig won't run? I see that regardless if one or the other is true in these if statements it will run the same handleProfilerRemoved method doesn't seem like that method makes a distinction between the two cases.

I understand i might be missing something, so maybe a comment here would be good to have in the case I am not understanding this correctly.

pkg/resource/training_job/custom_update.go
ananth102: correct

pkg/resource/training_job/hooks.go
ananth102: The profiler gets removed when ProfilerConfig or ProfilerRuleConfig (or both) are not present in the spec but were present before. 
>  it will run the same handleProfilerRemoved method doesn't seem like that method makes a distinction between the two cases

Correct, in either case the profiler gets disabled.

>However in this code if one of these updated it will return. So in this case the logic for ProfilerConfig won't run?

It only one gets updated(but not removed from the spec) the this method would return false and customSetInput would eventually get called.

I will document this in a comment

templates/training_job/sdk_update_post_set_output.go.tpl
ananth102: It gets the latest status after the update api has been called.

test/e2e/tests/test_trainingjob_debugger.py
```diff
@@ -142,16 +162,50 @@ def test_completed(self, xgboost_training_job_debugger):
...
+            training_sm_desc["LastModifiedTime"] != training_sm_desc["CreationTime"]
+        )
+
+    def delete_debugger_trainingjob(self, xgboost_training_job_debugger):
+        # Check that you can delete a completed resource from k8s
+        (reference, resource, _) = xgboost_training_job_debugger
+
+        training_job_name = resource["spec"].get("trainingJobName", None)
+
+        training_job_desc = get_sagemaker_training_job(training_job_name)
```
surajkota, line 197: you can just do `resource_arn = k8s.get_resource_arn(resource)` to get arn

surajkota (approved):
/approve

PR #190, opened 2023-01-12
fix: late initialize volumeInSizeGB field

ananth102:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/1628

Description of changes:
Adds volumeInSizeGB as a late initialized field in NotebookInstance.


pkg/resource/notebook_instance/manager.go
```diff
@@ -252,6 +252,9 @@ func (rm *resourceManager) incompleteLateInitialization(
 	if ko.Spec.PlatformIdentifier == nil {
 		return true
 	}
+	if ko.Spec.VolumeSizeInGB == nil {
```
surajkota, line 255: isn't nil returned by the service side as well? would that still result in incompleteLateInitialization?

ananth102: The service returns a non nil value, in the original issue the VolumeSizeInGB was not specified but the server returns 5. This causes the controller to try to update VolumeSizeInGB to be nil, which in turn causes the update loop.

PR #194, opened 2023-01-19
Resource retention Integration test

ananth102:
Issue #, if available:

Description of changes:
`test/e2e/resources/notebook_instance_lifecycle_retain.yaml` - Notebook Lifecycle CR with retain annotation
`test/e2e/tests/test_retain_policy.py` - Resource retention Integration test.


RedbackThomson (changes requested):
Hey @ananth102! Thanks for writing a test for this new feature! I'm always happy to see tests for this stuff coming out, so we can be sure that changes in runtime are not affecting service controllers. One question I have for you is why did you choose to apply this to the `NotebookInstanceLifecycle` resource? Is this just a relatively basic API that serves as a smoke test for the retain annotation?

ananth102:
Hi @RedbackThomson, we chose the NotebookInstanceLifecycle because it is a basic resource, this makes the test faster and also would also reduce the impact of any resource leakage. And yes, this test functions more as a smoke test.

test/e2e/tests/test_retain_policy.py
```diff
@@ -0,0 +1,105 @@
...
+from e2e import (
+    service_marker,
+    create_sagemaker_resource,
+    sagemaker_client,
+)
+
+
+from e2e.replacement_values import REPLACEMENT_VALUES
+
+RESOURCE_PLURAL = "notebookinstancelifecycleconfigs"
```
RedbackThomson, line 31: Even better would be to put this in the `test_notebook_instance_lifecycle_config.py` file and then import it in this file.

ananth102: Currently trying that, but it's making the Notebook lifecycle test slightly more flakey, need to investigate.

ananth102: Solved the issue.

test/e2e/tests/test_retain_policy.py
```diff
@@ -0,0 +1,106 @@
...
+    reference, spec, resource = create_sagemaker_resource(
+        resource_plural=RESOURCE_PLURAL,
+        resource_name=notebook_instance_lfc_name,
+        spec_file=RESOURCE_SPEC_FILE,
+        replacements=replacements,
+    )
+    assert resource is not None
+    yield (reference, resource, spec)
+
+    delete_notebook_instance_lifecycle_config(notebook_instance_lfc_name)
```
surajkota, line 49: assert on successful deletion

test/e2e/resources/notebook_instance_lifecycle_retain.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: sagemaker.services.k8s.aws/v1alpha1
+kind: NotebookInstanceLifecycleConfig
+metadata:
+  name: $NOTEBOOK_INSTANCE_LFC_NAME
+  annotations:
```
surajkota, line 5: not a big issue but it will be good to reuse existing spec since we are just adding an annotation

surajkota (changes requested):
Thanks, Great work on the feature Nick and Ananth for the test to ensure high bar

test/e2e/resources/notebook_instance_lifecycle_retain.yaml
ananth102: done

test/e2e/tests/test_retain_policy.py
```diff
@@ -0,0 +1,113 @@
...
+        # Delete the CR in Kubernetes
+        _, deleted = k8s.delete_custom_resource(reference)
+
+        assert deleted
+
+        # Verify that the resource was not deleted on SageMaker
+        assert (
+            get_notebook_instance_lifecycle_config(notebook_instance_lfc_name)
+            is not None
+        )
```
A-Hilaly, line 113: Maybe we should make a delete notebook (using the sdk) to make sure we don't have any dangling resources after?

ananth102: That is done here: https://github.com/aws-controllers-k8s/sagemaker-controller/blob/ed370d2b6aad386c657f5f65ef3b06868be366ef/test/e2e/tests/test_retain_policy.py#L51

ananth102:
\retest

PR #216, opened 2023-03-17
Refactor e2e deletion and unmark adopted Endpoint Test

ryansteakley:
Issue #, if available:

Description of changes:
Disabling the adopted endpoint test as canary since this does not provide value for a canary, will keep as e2e test. Refactors  
```
 _, deleted = k8s.delete_custom_resource(reference, cfg.JOB_DELETE_WAIT_PERIODS, cfg.JOB_DELETE_WAIT_LENGTH) 
> assert deleted 
```
into a shared method, and adds a retry loop for this assert.


test/e2e/__init__.py
```diff
@@ -372,3 +374,20 @@ def get_sagemaker_pipeline(pipeline_name: str):
             f"SageMaker could not find a pipeline with the name {pipeline_name}. Error {error}"
         )
         return None
+
+
+def try_delete_custom_resource(
```
surajkota, line 379: `k8s.delete_custom_resource` already has retries, do we need another wrapper method for retry logic?

max_retries = 2 same as changing cfg.DELETE_WAIT_PERIOD to 6 instead of 3? or increasing the period_length

wdyt

surajkota (changes requested):

test/e2e/__init__.py
ryansteakley: https://github.com/aws-controllers-k8s/test-infra/blob/0d6bfb1ae7539d9a92dd8a755e6d32ff33b900b0/src/acktest/k8s/resource.py#L196 it only retries on checking status, this will cause it to retry running the deletion command as-well.

test/e2e/__init__.py
```diff
@@ -372,3 +374,15 @@ def get_sagemaker_pipeline(pipeline_name: str):
...
         )
         return None
+
+
+def delete_custom_resource(
+    reference,
+    wait_period=cfg.JOB_DELETE_WAIT_PERIODS,
+    wait_length=cfg.JOB_DELETE_WAIT_LENGTH,
+):
+    deleted = True
```
surajkota, line 384: nit: default to false?

test/e2e/__init__.py
```diff
@@ -372,3 +374,15 @@ def get_sagemaker_pipeline(pipeline_name: str):
...
+
+    if k8s.get_resource_exists(reference): 
+        _, deleted = k8s.delete_custom_resource(reference, wait_period*2, wait_length)
```
surajkota, line 387: nit: lets double the value of the constant or at the passing method to keep the method generic 

lets use `cfg.DELETE_WAIT_PERIOD, cfg.DELETE_WAIT_LENGTH` for the ones with 3, 10? similar to job

surajkota (changes requested):

test/e2e/__init__.py
ryansteakley: Should default to true in case where resource does not exist so that assert passes. If resource existed the deletion method is called it will assign the true or false value as it normally would.

test/e2e/tests/test_model_package.py
```diff
@@ -93,11 +92,9 @@ def xgboost_versioned_model_package(xgboost_model_package_group):
...
     yield (reference, spec, resource)
     # Delete the k8s resource if not already deleted by tests
-    if k8s.get_resource_exists(reference):
-        _, deleted = k8s.delete_custom_resource(
-            reference, DELETE_WAIT_PERIOD, DELETE_WAIT_LENGTH
-        )
-        assert deleted
+
+
+assert delete_custom_resource(reference, DELETE_WAIT_PERIOD, DELETE_WAIT_LENGTH)
```
surajkota, line 97: incorrect indent

test/e2e/tests/test_model_package.py
```diff
@@ -116,11 +113,8 @@ def xgboost_unversioned_model_package():
 
     yield (reference, resource)
 
-    if k8s.get_resource_exists(reference):
-        _, deleted = k8s.delete_custom_resource(
-            reference, DELETE_WAIT_PERIOD, DELETE_WAIT_LENGTH
-        )
-        assert deleted
+
+assert delete_custom_resource(reference, DELETE_WAIT_PERIOD, DELETE_WAIT_LENGTH)
```
surajkota, line 117: incorrect indent

test/e2e/tests/test_notebook_instance.py
```diff
@@ -53,11 +54,9 @@ def notebook_instance():
...
 
     # Delete the k8s resource if not already deleted by tests
-    if k8s.get_resource_exists(reference):
-        _, deleted = k8s.delete_custom_resource(
-            reference, DELETE_WAIT_PERIOD, DELETE_WAIT_LENGTH
-        )
-        assert deleted
+
+
+assert delete_custom_resource(reference, DELETE_WAIT_PERIOD, DELETE_WAIT_LENGTH)
```
surajkota, line 59: incorrect indent

surajkota (changes requested):
few misses

surajkota (approved):
/approve
/lgtm

PR #230, opened 2023-03-31
Add select_regions_1 as config and unindent pytest command

ryansteakley:
Issue #, if available:

Description of changes:
Later refactor to this code, had incorrect indent which causes tests to not run when the ifstatement is true.


surajkota (approved):
/approve
/lgtm

PR #235, opened 2023-04-06
Increasing test timeouts and disabling stop tests

ananth102:
Issue #, if available:

Description of changes:

Monitoring Schedule can sometimes reach the scheduled state before the controller can verify that it is in Pending state. This PR removes the check for pending/resource synced.

Also increasing timeouts and disabling test_*_stopped tests

### DIFF
--- a/test/e2e/tests/test_monitoring_schedule.py
+++ b/test/e2e/tests/test_monitoring_schedule.py
@@ -165,19 +165,6 @@ class TestMonitoringSchedule:
         monitoring_schedule_arn = monitoring_schedule_desc["MonitoringScheduleArn"]
         assert k8s.get_resource_arn(resource) == monitoring_schedule_arn
 
-        # scheule transitions Pending -> Scheduled state
-        # Pending status is shortlived only for 30 seconds because baselining job has already been run
-        # remove the checks for Pending status if the test is flaky because of this
-        # as the main objective is to test for Scheduled status
-        # OR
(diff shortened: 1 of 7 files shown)


ryansteakley:
Why is the code change in model-monitor but everything else references model-package?

ananth102:
> Why is the code change in model-monitor but everything else references model-package?

My bad, resource names were too similar.

ryansteakley (approved):

test/e2e/tests/test_monitoring_schedule.py
```diff
@@ -165,19 +165,6 @@ def test_smoke(
...
-        # create the schedule with a on-going baseline job where it waits for the baselining job to complete
-        assert (
-            wait_resource_monitoring_schedule_status(
-                reference, self.STATUS_PENDING, 5, 2
-            )
-            == self.STATUS_PENDING
-        )
-        assert k8s.wait_on_condition(reference, "ACK.ResourceSynced", "False", 5, 2)
 
         self._assert_monitoring_schedule_status_in_sync(
```
surajkota, line 169: do we need to increase wait time for this check?

also can we do the same for feature group?

surajkota (changes requested):

surajkota (changes requested):
change for training as well - https://github.com/aws-controllers-k8s/sagemaker-controller/blob/45842199d81ed8b73f36dffe28396c269ae45ec5/test/e2e/tests/test_trainingjob.py#L64

lgtm otherwise

PR #240, opened 2023-04-13
Create shallow canary marker

rd-pong:
- Create shallow canary marker for eu-south-2, ap-southeast-3, me-central-1, eu-central-2, since some registries are not available in those regions.
- Add training and hosting to shallow canary
- Add placeholder values in replacement_values.py to avoid KeyError (not found)

Issue #, if available:

Description of changes:

### DIFF
--- a/test/e2e/tests/test_endpoint.py
+++ b/test/e2e/tests/test_endpoint.py
@@ -218,6 +218,7 @@ def faulty_config(name_suffix, single_container_model):
 
 
 @service_marker
+@pytest.mark.shallow_canary
 @pytest.mark.canary
 class TestEndpoint:
     def create_endpoint_test(self, xgboost_endpoint):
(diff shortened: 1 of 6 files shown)


test/e2e/tests/test_endpoint.py
```diff
@@ -218,6 +218,7 @@ def faulty_config(name_suffix, single_container_model):
 
 
 @service_marker
+@pytest.mark.shallow_canary
```
ryansteakley, line 221: This marker means if we run pytest with this marker then run this test. In your shell-script you are still running these tests with the regions in this `^(eu-south-2|ap-southeast-3|me-central-1|eu-central-2)` Correct if I'm wrong but isn't the goal of this pr to not run these tests in those regions since image is missing?

rd-pong: We want to run shallow canary, which includes training and hosting, in eu-south-2, ap-southeast-3, me-central-1, eu-central-2. Training and hosting does not require missing images such as xgboost, debugger and clarify.

rd-pong:
Verified  shallow canary if-else logic (run_test.sh) in CodeBuild. If service region is one of the four (eu-south-2, ap-southeast-3, me-central-1, eu-central-2), then run _test_endpoint_, _test_endpoint_config_ and _test_trainingjob_.

ryansteakley (approved):
/lgtm

PR #241, opened 2023-04-14
[Cherry-pick] Create shallow canary marker (#240)

rd-pong:
- Create shallow canary marker for eu-south-2, ap-southeast-3, me-central-1, eu-central-2, since some registries are not available in those regions.
- Add training and hosting to shallow canary
- Add placeholder values in replacement_values.py to avoid KeyError (not found)
- Add `--region ` for `aws s3 sync` to prevent error when downloading s3 bucket data to temp directory.
- Remove dependency on `s3://sagemaker-sample-data-$AWS_REGION `

### DIFF
--- a/test/e2e/replacement_values.py
+++ b/test/e2e/replacement_values.py
@@ -40,6 +40,13 @@ XGBOOST_IMAGE_URIS = {
     "eu-west-3": "659782779980.dkr.ecr.eu-west-3.amazonaws.com",
     "me-south-1": "801668240914.dkr.ecr.me-south-1.amazonaws.com",
     "sa-east-1": "737474898029.dkr.ecr.sa-east-1.amazonaws.com",
+    "eu-south-1": "978288397137.dkr.ecr.eu-south-1.amazonaws.com",
+    "ap-northeast-3": "867004704886.dkr.ecr.ap-northeast-3.amazonaws.com",
+    "eu-south-2": "104374241257.dkr.ecr.eu-south-2.amazonaws.com",
+    "ap-southeast-3": "951798379941.dkr.ecr.ap-southeast-3.amazonaws.com",
+    "me-central-1": "272398656194.dkr.ecr.me-central-1.amazonaws.com",
+    "eu-central-2": "680994064768.dkr.ecr.eu-central-2.amazonaws.com",
+    "af-south-1": "510948584623.dkr.ecr.af-south-1.amazonaws.com",
 }
 
 DEBUGGER_IMAGE_URIS = {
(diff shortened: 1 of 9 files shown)


ryansteakley (approved):

test/e2e/tests/test_endpoint_config.py
```diff
@@ -74,6 +74,7 @@ def single_variant_config():
 
 
 @service_marker
+@pytest.mark.shallow_canary
```
surajkota, line 77: can be removed

test/e2e/resources/kmeans_processingjob.yaml
```diff
@@ -21,7 +21,7 @@ spec:
   processingInputs:
     - inputName: mnist_tar
       s3Input:
-        s3URI: s3://sagemaker-sample-data-$AWS_REGION/algorithms/kmeans/mnist/mnist.pkl.gz
+        s3URI: s3://$SAGEMAKER_DATA_BUCKET/algorithms/kmeans/mnist/mnist.pkl.gz
```
surajkota, line 24: can you please move this under `$SAGEMAKER_DATA_BUCKET/sagemaker/processing/` in source bucket

test/e2e/replacement_values.py
```diff
@@ -63,6 +72,13 @@
     "eu-west-3": "447278800020.dkr.ecr.eu-west-3.amazonaws.com",
     "me-south-1": "986000313247.dkr.ecr.me-south-1.amazonaws.com",
     "sa-east-1": "818342061345.dkr.ecr.sa-east-1.amazonaws.com",
+    "eu-south-1": "563282790590.dkr.ecr.eu-south-1.amazonaws.com",
+    "ap-northeast-3": "479947661362.dkr.ecr.ap-northeast-3.amazonaws.com",
+    "af-south-1": "314341159256.dkr.ecr.af-south-1.amazonaws.com",
+    "eu-south-2": "",
```
surajkota, line 78: when does the key error occur? it means we cannot run a particular test in this region

shouldnt the test fail to find the image instead of this map returning an empty value?

test/e2e/service_bootstrap.py
```diff
@@ -55,6 +55,8 @@ def sync_data_bucket(bucket) -> str:
                 "sync",
                 f"s3://{SAGEMAKER_SOURCE_DATA_BUCKET}",
                 f"./{temp_dir}/",
+                "--region",
```
surajkota, line 58: curious, is this a new requirement? did you find anything in S3 docs which suggests this change in behavior?

surajkota (changes requested):

test/e2e/service_bootstrap.py
rd-pong: Experienced the following error in **some but not all** regions:
image (2)

My speculation is there is a mismatch between region we are calling the command and the one the bucket is in. [Explicitly specify](https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html) the region to use might be helpful. Some old regions might have more endpoints and can automatically switch, but for newer regions we need to explicitly specify.

References:
https://docs.aws.amazon.com/general/latest/gr/s3.html
https://repost.aws/questions/QUIZTCFPrzQXa2WJ2VaPC2lw/buckets-created-in-af-south-1-cannot-be-accessed-through-global-endpoint

test/e2e/tests/test_endpoint_config.py
rd-pong: Removed

test/e2e/resources/kmeans_processingjob.yaml
rd-pong: s3 URI changed, file moved to new location in S3

test/e2e/replacement_values.py
rd-pong: Removed placeholder values

PR #243, opened 2023-04-19
Add dummy values in replacement_values.py

rd-pong:
Issue #, if available:

Description of changes:
Add dummy values in replacement_values.py to avoid KeyError (when collecting tests)


surajkota (approved):
/approve
/lgtm

we can look at how to change the replacement_value implementation later so we dont need to do this for every new region

PR #261, opened 2024-03-01
Feature - Add/Modify/Delete tags for model resource

rkurduka:
Issue [1732](https://github.com/aws-controllers-k8s/community/issues/1732), if available: 

Description of changes:
 Currently tags modification is not supported for model resource . Hence added require changes to support tags modification. 

### DIFF
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -22,11 +22,20 @@ operations:
   StopPipelineExecution:
     operation_type: Delete
     resource_name: PipelineExecution
+  AddTags:
+    operation_type: Update
+    resource_name: Model
 resources:
   Model:
     hooks:
       delta_pre_compare:
         code: customSetDefaults(a, b)
+      sdk_read_one_post_set_output:
+        template_path: model/sdk_read_one_post_set_output.go.tpl
+      sdk_update_pre_build_request:
+        template_path: model/sdk_update_pre_build_request.go.tpl
+      sdk_update_post_build_request:
+        template_path: model/sdk_update_post_build_request.go.tpl
     exceptions:
       errors:
         404:
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -46,9 +46,9 @@ resources:
         - InvalidParameterValue
         - MissingParameter
     fields:
-      Tags:
-        compare:
-          is_ignored: true
+      # Tags:
+      #   compare:
+      #     is_ignored: true
       EnableNetworkIsolation:
         late_initialize:
           min_backoff_seconds: 5
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -30,6 +30,8 @@ resources:
     hooks:
       delta_pre_compare:
         code: customSetDefaults(a, b)
+      sdk_read_one_pre_set_output:
+        template_path: model/sdk_read_one_pre_set_output.go.tpl
       sdk_read_one_post_set_output:
         template_path: model/sdk_read_one_post_set_output.go.tpl
       sdk_update_pre_build_request:
--- a/config/crd/common/bases/services.k8s.aws_fieldexports.yaml
+++ b/config/crd/common/bases/services.k8s.aws_fieldexports.yaml
@@ -3,7 +3,8 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.14.0
+    controller-gen.kubebuilder.io/version: v0.9.2
+  creationTimestamp: null
   name: fieldexports.services.k8s.aws
 spec:
   group: services.k8s.aws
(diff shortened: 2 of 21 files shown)


apis/v1alpha1/generator.yaml
```diff
@@ -22,11 +22,22 @@ operations:
   StopPipelineExecution:
     operation_type: Delete
     resource_name: PipelineExecution
+  AddTags:
+    operation_type: Update
+    resource_name: Model
```
a-hilaly, line 27: This generates code that treats `AddTags` as the Update API Call to update the model.

helm/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: sagemaker-chart
 description: A Helm chart for the ACK service controller for Amazon SageMaker (SageMaker)
-version: 1.2.7
-appVersion: 1.2.7
+version: 0.0.0-non-release-version
```
a-hilaly, line 4: Please pull the controller tags

config/crd/common/bases/services.k8s.aws_fieldexports.yaml
```diff
@@ -3,7 +3,8 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.14.0
+    controller-gen.kubebuilder.io/version: v0.9.2
```
a-hilaly, line 6: You need to install controller-gen v0.14.0

templates/model/sdk_update_pre_build_request.go.tpl
```diff
@@ -0,0 +1,5 @@
+// this to handle delete/remove tags 
+_ , err = rm.deleteTags(ctx,desired,latest)
+if err != nil {
+	return nil, err
+}
```
a-hilaly, line 5: We don't need to delete the tags before performing a tag update.

pkg/resource/model/hooks.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"context"
+
+	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
+	svcapitypes "github.com/aws-controllers-k8s/sagemaker-controller/apis/v1alpha1"
+	svcsdk "github.com/aws/aws-sdk-go/service/sagemaker"
+)
+
+// deleteTags is used to keep tags in sync by calling Create and Delete API's
+func (rm *resourceManager) deleteTags(
```
a-hilaly, line 25: This function should stay, `syncTags` and also handle tag creation

a-hilaly (commented):
@rkurduka we have plans to start generating Tag handlers for all the ACK controllers. But nothing stops us from implement some manually if there is urgent need for them :)
I left few comments in-line

rkurduka:
> @rkurduka we have plans to start generating Tag handlers for all the ACK controllers. But nothing stops us from implement some manually if there is urgent need for them :) I left few comments in-line

@a-hilaly  - thanks for update . Yes there was customer ask for this . Issue 1732 has details

apis/v1alpha1/generator.yaml
rkurduka: @a-hilaly - Thanks for review.
Sagemaker model API do not have any modify /update method , as fields are immutable except tags . Also Sagemaker API resource only have "ADDTAG" and "DELETE TAG" . So to handle update on this resource will have to use one of these two. So chose "ADD TAGS" and handling "DELETE tags" using template hooks. 

Let me know if you think any other approach we can consider

helm/Chart.yaml
rkurduka: Done

config/crd/common/bases/services.k8s.aws_fieldexports.yaml
rkurduka: Done

templates/model/sdk_update_pre_build_request.go.tpl
rkurduka: @a-hilaly  - as I mentioned above , SDK update method is handling ADDTAGS only  and hooks are deleting tags , there is no modify API available for sagemaker model resource . So in SDKUPDATE , first "DeletTags" method look for any deletion request , if not then proceed and look for add tags request .

Let me know if anything else is require.

pkg/resource/model/hooks.go
rkurduka: @a-hilaly  - same comments as above , let me know if any other approach we should consider here

pkg/resource/model/hooks.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	if c.Value != nil {
+		res.SetValue(*c.Value)
+	}
+
+	return res
+}
+
+// computeTagsDelta returns tags to be added and removed from the resource
+func computeTagsDelta(
```
surajkota, line 96: Sagemaker tag API pattern(addTag/deleteTag) resembles memorydb pattern ([TagResource](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/memorydb/client/tag_resource.html)/UntagResource). 

We have a reference implementation in memorydb which uses `ackcompare.GetTagsDifference` helper method which can reduce maintenance of SM controller code. https://github.com/aws-controllers-k8s/memorydb-controller/blob/main/pkg/resource/acl/hooks.go#L92

Can you please look into adopting similar pattern?

FYI @ryansteakley @a-hilaly

surajkota (changes requested):
See https://github.com/aws-controllers-k8s/sagemaker-controller/pull/261#discussion_r1513162375

apis/v1alpha1/generator.yaml
surajkota: Agree with Amine, need to discuss this further as we need to throw appropriate error if any other fields except Tags are modified as Model does not have udpate API, maybe implement it completely via hooks

pkg/resource/model/hooks.go
rkurduka: Thanks @surajkota for sharing this , and looking into it . But it seems memoryDB - ACL resource do have "UpdateACL" API method . So we can plug tags manipulation in it , using hooks. But in sagemaker -model case , there is NO update method available , so "SDKUPDATE block doesn't get generated by code-generator . So had to generate that using "ADDTags" in "Operation" Block , and later handling tags CRUD operations in it . Let me know if you have any suggestion in this case. Thanks again

apis/v1alpha1/generator.yaml
rkurduka: @surajkota - Added check in update_pre hooks to validate field to update, if it is other than tag , reconciler will update "message" status.

Please let me know if you feel anything else is require

PR #275, opened 2024-05-22
Bugfix - handled panic due to difference in InputdataConfig size 

ankurjha7:
Description of changes:

The change handles panic due to **runtime error : index out of range**  . The present code accesses b.ko.Spec.InputDataConfig[index] which can throw panic if size of b.ko.Spec.InputDataConfig is smaller than a.ko.Spec.InputDataConfig 

The solution checks for this condition and sets the a.ko.Spec.InputDataConfig[index].RecordWrapperType to default type 


a-hilaly:
cc @aws-controllers-k8s/sagemaker-maintainer

a-hilaly:
cc @ryansteakley

PR #276, opened 2024-07-10
test(inference-component): Remove update check

ananth102:
Test is not working

Issue #, if available:

Description of changes:
- Canary is failing.

### DIFF
--- a/test/e2e/tests/test_inference_component.py
+++ b/test/e2e/tests/test_inference_component.py
@@ -245,17 +245,6 @@ class TestInferenceComponent:
         resource = k8s.wait_resource_consumed_by_controller(reference)
         assert resource is not None
 
-        # inference component transitions Updating -> InService state
-        assert_inference_component_status_in_sync(
-            reference.name,
-            reference,
-            cfg.INFERENCE_COMPONENT_STATUS_UPDATING,
-        )
-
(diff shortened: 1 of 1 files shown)


test/e2e/tests/test_inference_component.py
```diff
@@ -245,17 +245,6 @@ def update_inference_component_failed_test(self, inference_component, faulty_mod
...
-        assert_inference_component_status_in_sync(
-            reference.name,
-            reference,
-            cfg.INFERENCE_COMPONENT_STATUS_UPDATING,
-        )
-
-        assert k8s.wait_on_condition(reference, "ACK.ResourceSynced", "False")
-        assert k8s.get_resource_condition(reference, "ACK.Terminal") is None
-        resource = k8s.get_resource(reference)
-
```
a-hilaly, line 258: Can you elaborate more why those tests are failing now? did something change from SM side, or are we dealing with a flaky test?

ananth102: > Can you elaborate more why those tests are failing now?

Update takes ~10 seconds, it used to take 2 minutes. This causes the check here to fail as the Inference Component will be in InService status and not Updating status. The failure reason is also different than before.
But even the other update test is failing per the CI so there may be a bigger issue.

> did something change from SM side, or are we dealing with a flaky test?

Not a flaky test this has been consistently failing for the last week.

PR #345, opened 2026-01-22
Fix Domain stuck late initializing

knottnt:
Issue #, if available:

Description of changes:
After initially creating a Domain the resource could get stuck in a loop attempting to late initialize DefaultUserSettings.SpaceStorageSettings as this field is not always set. Additionally, the controller could see erroneous deltas due to some DefaultUserSettings top level configs returning an empty object after an update was applied to the Domain.

- Apply `skip_incomplete_check` to DefaultUserSettings.SpaceStorageSettings to avoid retrying late initialization if not set in Service API
- Late initialize top level DefaultUserSettings configs with `skip_incomplete_check` set to pickup changes from null to an empty object.
- Add checks verifying Domain resource has synced condition to e2e tests 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.swp
 *~
 .idea
+.vscode
 /docs/site
 bin
 test/bin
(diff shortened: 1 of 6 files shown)


michaelhtm (approved):
👍 
/lgtm

